PR #1, opened 2021-11-04
Add packages to SUMMARY

tzakian:
This PR adds packages to the summary for the Move book -- the package should show up from this.

I generated this by running `mdbook build` in the `language/documentation/book` directory in the Diem repo, and then copying the contents of the generated "book" directory here. Let me know if this is the incorrect generation process!


bob-wilson:
Can you regenerate from before #9348? Todd made some changes for assert that are not present in the current (1.5) release branch.

tzakian:
Should be regenerated from just before #9348 now. Let me know if it looks good

bob-wilson (approved):
Looks good

PR #2, opened 2021-12-03
Update Move book off of head, add unit testing section

tzakian:
This PR adds a section to the Move book on unit testing, and also updates the book to be built off of `diem/main` -- so picks up the new syntax changes to `assert!`


bob-wilson:
We had previously tried to keep the published docs aligned with the latest release. Are you proposing to change that in general or is this a one-off change to temporarily sync up to a more recent version?

tzakian:
Longer-term, I'd propose that that we change it in general until we have Move releases given the recent developments. Near-term I'm happy to hold-off on the larger question, and have this be a one-off change  for the hackathon since they will be building off of `main`, and it would be good for the book to have the updated syntax/docs for that along with docs on unit testing.

bob-wilson:
OK, I'm fine with it as a one-off. In general, though, we should work toward a more consistent answer about whether people should use the main branch or the latest release.

bob-wilson (approved):
I skimmed through this and the changes do indeed look like they all match up with the latest content from the main branch.

PR #13, opened 2022-01-14
[move-package] Make package system thread/process safe

tzakian:
This PR adds a global package lock to the Move package manager: we have both a thread-local mutex and cross-process locking. The issue that this PR resolves was especially apparent when dealing with git dependencies. This PR also adds two tests to make sure that these locks are being respected by threads and processes and solve the issues reported in https://github.com/diem/diem/issues/10102.

### DIFF
--- a/language/tools/move-package/src/lib.rs
+++ b/language/tools/move-package/src/lib.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 pub mod compilation;
+mod package_lock;
 pub mod resolution;
 pub mod source_package;
 
@@ -22,6 +23,7 @@ use crate::{
     compilation::{
         build_plan::BuildPlan, compiled_package::CompiledPackage, model_builder::ModelBuilder,
     },
+    package_lock::PackageLock,
     resolution::resolution_graph::{ResolutionGraph, ResolvedGraph},
     source_package::{layout, manifest_parser},
 };
@@ -101,7 +103,10 @@ impl BuildConfig {
         writer: &mut W,
     ) -> Result<(CompiledPackage, CompilationCachingStatus)> {
         let resolved_graph = self.resolution_graph_for_package(path)?;
-        BuildPlan::create(resolved_graph)?.compile(writer)
+        let mutx = PackageLock::lock();
+        let ret = BuildPlan::create(resolved_graph)?.compile(writer);
+        mutx.unlock();
+        ret
     }
 
     // NOTE: If there are no renamings, then the root package has the global resolution of all named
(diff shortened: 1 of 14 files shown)


language/tools/move-package/src/lib.rs
```diff
@@ -115,10 +120,14 @@ impl BuildConfig {
...
         let resolved_graph = self.resolution_graph_for_package(path)?;
-        ModelBuilder::create(resolved_graph, model_config).build_model()
+        let mutx = PackageLock::lock();
+        let ret = ModelBuilder::create(resolved_graph, model_config).build_model();
+        mutx.unlock();
+        ret
     }
 
     pub fn resolution_graph_for_package(mut self, path: &Path) -> Result<ResolvedGraph> {
+        let mutx = PackageLock::lock();
```
sblackshear, line 130: Not sure I immediately follow the granularity of this lock. I would intuitively expect it to go before 139 or 140, but I don't have a good understanding of which paths might be touched in other threads. Could be worth a comment.

language/tools/move-package/src/package_lock.rs
```diff
@@ -0,0 +1,37 @@
...
+use named_lock::{NamedLock, NamedLockGuard};
+use once_cell::sync::Lazy;
+use std::sync::{Mutex, MutexGuard};
+
+const PACKAGE_LOCK_NAME: &str = "move_pkg_lock";
+static PACKAGE_THREAD_MUTEX: Lazy<Mutex<()>> = Lazy::new(|| Mutex::new(()));
+static PACKAGE_PROCESS_MUTEX: Lazy<NamedLock> =
+    Lazy::new(|| NamedLock::create(PACKAGE_LOCK_NAME).unwrap());
+
+// The package lock is a lock held across threads and processes
```
sblackshear, line 13: Probably useful to explain what the package lock is protecting in some detail (I think the global package cache and its subdirectories) + when it needs to be acquired.

sblackshear (approved):
Thanks so much for this Tim!! Everything makes sense to me.

language/tools/move-package/src/lib.rs
tzakian: Good question, looking at it since this take ownership of `self` the lock only needs to be around 139/140, but I figured to be on the safe side I'd lock the whole body. I'll update

sblackshear: I'm not worried about the perf of a coarser grained lock here, but I think placing them very precisely is good for understandability.

sblackshear (approved):
Re-stamping under the assumption that this repo also requires it, let me know if I'm wrong about that.

PR #14, opened 2022-01-14
[move-prover] Print out global memory in the counterexample trace

rahxephon89:
## Motivation

A new feature to print out the content of the related global memory along with the counter example trace when "--trace" is set in the prover. In this PR, we define the global memory as the memory that is accessed in the function where the verification fails.

The output looks like:

```
    =         Resource name: DiemTimestamp_CurrentTimeMicroseconds 
    =         Values: 
    =           {
    =             Address(173345816): DiemTimestamp.CurrentTimeMicroseconds{
    =               microseconds = 18446744073709551615},
    =             Default: DiemTimestamp.CurrentTimeMicroseconds{
    =               microseconds = 18446744073709551615}}
```

Yes

## Test Plan

cargo test


wrwg (approved):
I think we should move forward with this. We can iterate as needed.

wrwg (approved):

PR #16, opened 2022-01-14
[move-package] Add checks for case insensitive file systems

tzakian:
This PR adds a check to the Move package manager that checks that two source files will not result in a conflict if being run on case insensitive file systems.

We take the view that all filesystems are case insensitive to (1) simplify the logic (since we don't need to determine if the file system is case sensitive or not); and (2) will help maximize portability of packages -- Bob can develop on Linux and feel confident that Alice can use that package on Mac without issues.

Care has been put into the error messages for this error case to provide help on how to solve the problem if it occurs during compilation. You can see an example of the new error message in the test that is added in this PR.


language/tools/move-package/src/compilation/compiled_package.rs
```diff
@@ -652,7 +652,65 @@ impl CompiledPackage {
...
+        for compiled_unit in &self.compiled_units {
+            let is_module = matches!(&compiled_unit.unit, CompiledUnit::Module(_));
+            let name = match &compiled_unit.unit {
+                CompiledUnit::Script(named) => named.name.as_str(),
+                CompiledUnit::Module(named) => named.name.as_str(),
+            };
+            let entry = insensitive_mapping
+                .entry(name.to_lowercase())
+                .or_insert_with(BTreeMap::new);
+            let insert = entry.entry(name).or_insert_with(Vec::new);
```
tnowacki, line 669: This is a weird name. I don't understand what is going on here because it is called `insert` I would expect this to be some plural noun.

language/tools/move-package/src/compilation/compiled_package.rs
```diff
@@ -652,7 +652,65 @@ impl CompiledPackage {
...
+            insert.push((
+                is_module,
+                compiled_unit.source_path.to_string_lossy().to_string(),
+            ));
+        }
+        let errs = insensitive_mapping
+            .into_iter()
+            .filter_map(|(insensitive_name, preimages)| {
```
tnowacki, line 677: I am confused why it is called preimages?

language/tools/move-package/src/compilation/compiled_package.rs
```diff
@@ -652,7 +652,65 @@ impl CompiledPackage {
         Ok((compiled_package, CompilationCachingStatus::Recompiled))
     }
 
+    // We take the (restrictive) view that all filesystems are case insensitive to maximize
+    // portability of packages.
+    fn check_filepaths_ok(&self) -> Result<()> {
+        // A mapping of (lowercase_name => (real_name, [info_for_each_occurence])
+        let mut insensitive_mapping = BTreeMap::new();
```
tnowacki, line 659: Thoughts about a slightly different mapping here? Might make this a bit easier to follow, but I could be wrong.
Just doing `lowercase_file_name => [info_for_each_occurrence]` I think it would make the logic in `errs` seem a bit less wonky for me

language/tools/move-package/src/compilation/compiled_package.rs
tzakian: Ah I was thinking of these as being the preimage of `insensitive_name` (i.e., all the names that mapped to it under `to_lowercase`). I'll update with a better name.

language/tools/move-package/src/compilation/compiled_package.rs
tzakian: Totally right. I got it into my head that I needed to group by case-sensitive name, but in reality I don't as long as `[info_for_each_occurence]` is allowed to not be a set. Will update :)

tnowacki (approved):
I might poke around at this code to see if it can be done differently

But for now lets land this! thanks!

PR #20, opened 2022-01-19
[bytecode verifier] Improve reference safety performance

tnowacki:
- Improve CFG traversal for bytecode verifier
- Improve borrow graph performance

## Test Plan

- Ran tests
- Some rough benchmarks

### DIFF
--- a/language/move-binary-format/src/control_flow_graph.rs
+++ b/language/move-binary-format/src/control_flow_graph.rs
@@ -22,6 +22,8 @@ pub trait ControlFlowGraph {
     /// Successors of the block ID in the bytecode vector
     fn successors(&self, block_id: BlockId) -> &Vec<BlockId>;
 
+    fn next_block(&self, block_id: BlockId) -> Option<CodeOffset>;
+
     /// Iterator over the indexes of instructions in this block
     fn instr_indexes(&self, block_id: BlockId) -> Box<dyn Iterator<Item = CodeOffset>>;
 
@@ -34,10 +36,16 @@ pub trait ControlFlowGraph {
     /// Return the id of the entry block for this control-flow graph
     /// Note: even a CFG with no instructions has an (empty) entry block.
     fn entry_block_id(&self) -> BlockId;
+
+    /// Checks if the block ID is a loop head
+    fn is_loop_head(&self, block_id: BlockId) -> bool;
+
+    /// Checks if the the edge from cur->next is a back edge
+    /// returns false if the edge is not in the cfg
+    fn is_back_edge(&self, cur: BlockId, next: BlockId) -> bool;
 }
 
 struct BasicBlock {
-    entry: CodeOffset,
     exit: CodeOffset,
     successors: Vec<BlockId>,
 }
@@ -46,12 +54,16 @@ struct BasicBlock {
 pub struct VMControlFlowGraph {
     /// The basic blocks
     blocks: Map<BlockId, BasicBlock>,
+    /// Basic block ordering for traversal
+    traversal_successors: Map<BlockId, BlockId>,
+    /// Map of loop heads with all of their back edges
+    loop_heads: Map<BlockId, /* back edges */ Set<BlockId>>,
 }
 
 impl BasicBlock {
-    pub fn display(&self) {
+    pub fn display(&self, entry: BlockId) {
         println!("+=======================+");
-        println!("| Enter:  {}            |", self.entry);
+        println!("| Enter:  {}            |", entry);
         println!("+-----------------------+");
         println!("==> Children: {:?}", self.successors);
         println!("+-----------------------+");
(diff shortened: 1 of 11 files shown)


language/move-binary-format/src/control_flow_graph.rs
```diff
@@ -73,32 +93,104 @@ impl VMControlFlowGraph {
...
+                let non_loop_continue_successors = block
+                    .successors
+                    .iter()
+                    // remove the loop back edges
+                    .filter(|successor| **successor > id)
+                    .copied()
+                    .collect();
+                (id, non_loop_continue_successors)
+            })
+            .collect();
```
sblackshear, line 130: Some `debug_assert!`-guarded code that checks that this is actually a DAG would make me feel warmer/fuzzier...

language/move-binary-format/src/control_flow_graph.rs
```diff
@@ -46,12 +50,27 @@ struct BasicBlock {
 pub struct VMControlFlowGraph {
     /// The basic blocks
     blocks: Map<BlockId, BasicBlock>,
+    /// Basic block ordering for traversal
+    traversal_successors: Map<BlockId, BlockId>,
+    /// Map from blocks with a loop's last continue to that loop's head
+    /// "last" here meaning at the highest bytecode offset
+    /// "continue" simply being a back-edge, but it is a continue in the sense that the loop
+    /// back-edge resembles a "continue" in its control flow layout
+    loop_last_continue_blocks: Map<BlockId, BlockId>,
```
sblackshear, line 59: What is the difference between a "loop last continue" and a back edge? It sounds like all loop last continues are back edges, but not the other way around. Are there other differences? Basically, I'm trying to figure out why we use these and not back edges...

language/move-binary-format/src/control_flow_graph.rs
```diff
@@ -73,32 +93,104 @@ impl VMControlFlowGraph {
...
+        // build a DAG subgraph (remove the loop back edges)
+        let dag: Map<BlockId, Set<BlockId>> = blocks
+            .iter()
+            .map(|(id, block)| {
+                let id = *id;
+                let non_loop_continue_successors = block
+                    .successors
+                    .iter()
+                    // remove the loop back edges
+                    .filter(|successor| **successor > id)
```
sblackshear, line 125: Is this sufficient to check that it's a back edge? I would think in principal we could have a CFG like

```
B1
  goto B3;
B2:
  return
B3:
  goto B2;
```

I think this code will treat B3 -> B2 as a loop back edge, but it is not.

language/move-binary-format/src/control_flow_graph.rs
tnowacki: It's the last back edge. But maybe you are right and you can just use the first back edge? In some ways, I think this might be an over optimization?

A loop here is a series of blocks. There might be many back edges in that set of blocks to the loop head. With this there are two options:
- Revisit the loop in the absint traversal when any backedge changes the pre state of the loop head's block 
- Revisit the loop in the absint traversal once all blocks with the loop head's block as a successor are visited

Setting up this "last continue" nonsense takes the latter approach. You only revisit the loop head block once the entirety of the loop has been visited at least once

language/move-binary-format/src/control_flow_graph.rs
tnowacki: That code would not be allowed in our current CFG set of checks (see language/move-bytecode-verifier/src/control_flow.rs). This would be counted as "jumping into the middle of a loop"

sblackshear: Can this check now do something like `!is_back_edge(successor, id)` instead of `**successor > id`

language/move-binary-format/src/control_flow_graph.rs
sblackshear: Thank you, I feel warmer/fuzzier.

language/move-borrow-graph/src/references.rs
```diff
@@ -44,10 +44,16 @@ pub(crate) struct BorrowEdge<Loc: Copy, Lbl: Clone + Ord> {
     pub(crate) loc: Loc,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct BorrowEdgeSet<Loc: Copy, Lbl: Clone + Ord> {
+    edges: BTreeSet<BorrowEdge<Loc, Lbl>>,
+    overflown: bool,
```
sblackshear, line 50: I think this needs some explanation

tnowacki: lol I was trying not too at the time, but sure I'll add a high-level one

sblackshear (dismissed):

PR #21, opened 2022-01-19
[move-compiler] Improve CFG traversal

tnowacki:
- Use reverse post-order traversal for source language CFG
- Will allow for warnings in the future on type-system checks that require fix-point iteration

## Motivation

- We want to change the rules of the bytecode verifier to not require fix-point iteration for its rules
- This refactor will allow for these sorts of constraints to exist for the source language too

## Test Plan

- internal only change
- ran the tests

### DIFF
--- a/language/move-compiler/src/cfgir/absint.rs
+++ b/language/move-compiler/src/cfgir/absint.rs
@@ -87,59 +87,63 @@ pub trait AbstractInterpreter: TransferFunctions {
     ) -> (BTreeMap<Label, Self::State>, Diagnostics) {
         let mut inv_map: InvariantMap<Self::State> = InvariantMap::new();
         let start = cfg.start_block();
-        let mut work_list = vec![start];
-        inv_map.insert(
-            start,
-            BlockInvariant {
-                pre: initial_state,
-                post: BlockPostcondition::Unprocessed,
-            },
-        );
+        let mut next_block = Some(start);
 
-        while let Some(block_lbl) = work_list.pop() {
-            let block_invariant = match inv_map.get_mut(&block_lbl) {
-                Some(invariant) => invariant,
-                None => unreachable!("Missing invariant for block"),
-            };
+        while let Some(block_label) = next_block {
+            let block_invariant = inv_map
+                .entry(block_label)
+                .or_insert_with(|| BlockInvariant {
+                    pre: initial_state.clone(),
+                    post: BlockPostcondition::Unprocessed,
+                });
 
-            let (post_state, errors) = self.execute_block(cfg, &block_invariant.pre, block_lbl);
+            let (post_state, errors) = self.execute_block(cfg, &block_invariant.pre, block_label);
             block_invariant.post = if errors.is_empty() {
                 BlockPostcondition::Success
             } else {
                 BlockPostcondition::Error(errors)
             };
+
             // propagate postcondition of this block to successor blocks
-            for next_lbl in cfg.successors(block_lbl) {
-                match inv_map.get_mut(next_lbl) {
+            let mut next_block_candidate = cfg.next_block(block_label);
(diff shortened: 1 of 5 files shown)


vgao1996 (approved):
Clever optimization! Let's get it in...

language/move-compiler/src/cfgir/cfg.rs
```diff
@@ -347,16 +397,111 @@ fn maybe_unmark_infinite_loop_starts(
...
+    all_labels: impl IntoIterator<Item = Label>,
+    successor_map: &BTreeMap<Label, BTreeSet<Label>>,
+    include_dead_code: bool,
+) -> (
+    /* order */ Vec<Label>,
+    /* last continues */ BTreeMap<Label, Label>,
+) {
+    // Determine traversal order
+    // build a DAG subgraph (remove the loop back edges)
+    let dag: BTreeMap<Label, BTreeSet<Label>> = successor_map
```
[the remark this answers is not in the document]
tnowacki: Good Call! But I figured the set wouldn't be that big, so I'd rather just do the set than have to explain why the vec is in fact a set

language/move-compiler/src/cfgir/absint.rs
```diff
@@ -87,59 +87,69 @@ pub trait AbstractInterpreter: TransferFunctions {
...
                 BlockPostcondition::Success
             } else {
                 BlockPostcondition::Error(errors)
             };
+
             // propagate postcondition of this block to successor blocks
-            for next_lbl in cfg.successors(block_lbl) {
-                match inv_map.get_mut(next_lbl) {
+            for next_block_id in cfg.successors(block_label) {
+                match inv_map.get_mut(next_block_id) {
```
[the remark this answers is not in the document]
tnowacki: I think that would make the code generally more complicated. But maybe I'm not seeing how it would work out in its entirety

language/move-compiler/src/cfgir/cfg.rs
```diff
@@ -25,6 +28,12 @@ pub trait CFG {
     fn commands<'a>(&'a self, label: Label) -> Box<dyn Iterator<Item = (usize, &'a Command)> + 'a>;
     fn num_blocks(&self) -> usize;
     fn start_block(&self) -> Label;
+
+    fn next_block(&self, label: Label) -> Option<Label>;
+
+    fn last_continue_to_loop(&self, label: Label) -> Option<Label>;
```
sblackshear, line 34: Can you formally to define this `last_continue_to_loop` concept in terms of nodes and edges in the CFG (and perhaps reify this explanation in a comment?). I think I asked this elsewhere--apologies if I missed the answer. But I think it's pretty important for someone trying to read/modify this code to understand the concept in depth.

tnowacki: Would this be an appropriate enough comment:
"Blocks have their labels generated in ascending order based on the position in the original source text. This means that blocks with higher numerically valued labels appear later in the CFG than those with lower valued labels. The "last continue" of the loop is the back edge with the highest numerical value. In other words, it represents the "end" of the loop as the loop appeared in the source text (and in turn in the CFG)

language/move-compiler/src/cfgir/cfg.rs
```diff
@@ -524,6 +744,25 @@ fn ast_debug_cfg<'a>(
...
     });
 
+    w.write("traversal:");
+    w.indent(4, |w| {
+        for (cur, next) in traversal {
+            w.writeln(&format!("{} => {}", cur, next))
+        }
+    });
+
+    w.write("loop last continues:");
```
sblackshear, line 754:
```suggestion
    w.write("loop heads:");
```

sblackshear (dismissed):

vgao1996 (approved):

PR #31, opened 2022-01-29
handle errors in execute_function_for_effects()

stella3d:
## Motivation

`.unwrap()` calls cause panics, returning the error to the user instead is much more useful.

Yes

## Test Plan

I've added tests for some of the new error handling: 
* when ref `&` and mutable ref `&mut` arguments fail to deserialize

Testing error handling for written mutable ref values that fail to serialize is less obvious to me, but i'd like to have a test for it as well.

### DIFF
--- a/language/move-vm/integration-tests/src/tests/exec_func_effects_tests.rs
+++ b/language/move-vm/integration-tests/src/tests/exec_func_effects_tests.rs
@@ -7,14 +7,14 @@ use move_core_types::{
     account_address::AccountAddress,
     identifier::Identifier,
     language_storage::ModuleId,
-    value::{MoveValue, serialize_values}, vm_status::StatusCode
+    value::{serialize_values, MoveValue},
+    vm_status::StatusCode,
 };
 use move_vm_runtime::{move_vm::MoveVM, session::ExecutionResult};
 use move_vm_test_utils::InMemoryStorage;
 use move_vm_types::gas_schedule::GasStatus;
 
-use crate::compiler::{compile_units, as_module};
-
+use crate::compiler::{as_module, compile_units};
 
 const TEST_ADDR: AccountAddress = AccountAddress::new([42; AccountAddress::LENGTH]);
 const TEST_MODULE_ID: &str = "M";
@@ -28,19 +28,26 @@ const FUN_NAMES: [&str; 2] = [USE_MUTREF_LABEL, USE_REF_LABEL];
 fn fail_arg_deserialize() {
     let mod_code = setup_module();
     // all of these should fail to deserialize because the functions expect u64 args
-    vec![MoveValue::U8(16), MoveValue::U128(512), MoveValue::Bool(true)]
+    vec![
+        MoveValue::U8(16),
+        MoveValue::U128(512),
+        MoveValue::Bool(true),
+    ]
     .iter()
     .for_each(|mv| {
-        FUN_NAMES.iter().for_each(|name| {
-            match run(&mod_code, name, mv.clone()) {
+        FUN_NAMES
+            .iter()
+            .for_each(|name| match run(&mod_code, name, mv.clone()) {
                 ExecutionResult::Success { .. } => {
                     panic!("Should have failed to deserialize non-u64 type to u64");
-                },
-                ExecutionResult::Fail { error, .. } => {
-                    assert_eq!(error.major_status(), StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT);
                 }
-            }
-        });
+                ExecutionResult::Fail { error, .. } => {
+                    assert_eq!(
+                        error.major_status(),
+                        StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT
+                    );
+                }
+            });
     });
 }
 
@@ -49,11 +56,13 @@ fn fail_arg_deserialize() {
 fn mutref_output_success() {
     let mod_code = setup_module();
     match run(&mod_code, USE_MUTREF_LABEL, MoveValue::U64(1)) {
-        ExecutionResult::Success { mutable_ref_values, .. } => {
+        ExecutionResult::Success {
+            mutable_ref_values, ..
+        } => {
             assert_eq!(1, mutable_ref_values.len());
             let parsed = parse_u64_arg(mutable_ref_values.first().unwrap());
             assert_eq!(EXPECT_MUTREF_OUT_VALUE, parsed);
-        },
+        }
         ExecutionResult::Fail { error, .. } => {
             panic!("{:?}", error);
         }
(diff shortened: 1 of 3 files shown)


stella3d:
Tried to format the errors similarly to the other code around this function.  

`arg_to_actual()` could maybe have a clearer name.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -377,19 +377,22 @@ impl VMRuntime {
         for (idx, (arg, arg_type)) in args.into_iter().zip(params).enumerate() {
             match arg_type {
                 Type::MutableReference(inner_t) => {
-                    dummy_locals
-                        .store_loc(idx, self.deserialize_value(&inner_t, arg).unwrap())
-                        .unwrap();
-                    actuals.push(dummy_locals.borrow_loc(idx).unwrap());
+                    if let Err(err) = self.arg_to_actual(idx, arg, &inner_t, &mut dummy_locals, &mut actuals) {
```
wrwg, line 380: Can't one use `self.arg_to_actual(...).map_err(..)?` or something?

language/move-vm/runtime/src/runtime.rs
```diff
@@ -377,19 +377,22 @@ impl VMRuntime {
...
-                    dummy_locals
-                        .store_loc(idx, self.deserialize_value(&inner_t, arg).unwrap())
-                        .unwrap();
-                    actuals.push(dummy_locals.borrow_loc(idx).unwrap())
+                    if let Err(err) = self.arg_to_actual(idx, arg, &inner_t, &mut dummy_locals, &mut actuals) {
+                        return Err(err.finish(Location::Undefined));
+                    }
+                }
+                arg_type => {
+                    match self.deserialize_value(&arg_type, arg) {
```
wrwg, line 391: Similarly here.

language/move-vm/runtime/src/runtime.rs
```diff
@@ -432,18 +435,41 @@ impl VMRuntime {
...
+                None => return Err(PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR)
+                                .with_message("failed to serialize mutable ref values".to_string())
+                                .finish(Location::Undefined)),
+            };
         }
 
         Ok((serialized_return_vals, serialized_mut_ref_outputs))
     }
+    
+    fn arg_to_actual(&self, idx: usize, arg: Vec<u8>, arg_t: &Box<Type>, locals: &mut Locals, actuals: &mut Vec<Value>)
```
sblackshear, line 458: > arg_to_actual() could maybe have a clearer name.

Maybe:
```
fn borrow_arg(&self, idx: usize, arg: Vec<u8>, arg_t: &Box<Type>, locals: &mut Locals) -> Value
```
 , change the function to return the result of `locals.borrow_loc(idx)`, and do the `actuals.push` at the call sites?

language/move-vm/integration-tests/src/tests/exec_func_effects_tests.rs
```diff
@@ -0,0 +1,126 @@
...
+                },
+                ExecutionResult::Fail { error, .. } => {
+                    assert_eq!(error.major_status(), StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT);
+                }
+            }
+        });
+    });
+}
+
+// check happy path for writing to mut ref args - may be unecessary / covered by other tests
```
stella3d, line 47: What i really want to test is the new error handling for mutable ref serialization, but it seems tricky to get the Move code to return a value that fails to serialize. If I can figure out how to cause that error, i'll add the test for it to mirror this.

language/move-vm/integration-tests/src/tests/exec_func_effects_tests.rs
```diff
@@ -0,0 +1,126 @@
...
+        }}
+    "#,
+        TEST_ADDR, TEST_MODULE_ID, USE_MUTREF_LABEL, EXPECT_MUTREF_OUT_VALUE, USE_REF_LABEL
+    );
+
+    let module_id = ModuleId::new(TEST_ADDR, Identifier::new(TEST_MODULE_ID).unwrap());
+    (module_id, code)
+}
+
+fn run(module: &ModuleCode, fun_name: &str, arg_val0: MoveValue) -> ExecutionResult {
```
stella3d, line 82: the setup code is mostly adapted from other integration tests - the existing abstractions / utilities for tests didn't seem unified, so i've started trying to do that here.
Ideally i'd like to move most of this helper code below here out of this file and into shared lib.

language/move-vm/runtime/src/runtime.rs
stella3d: `map_err` can be used here, i tried reworking it a bit to use it but it didn't seem any shorter or more readable. Maybe there is something i'm missing in its use, do you have something specific in mind ?

language/move-vm/runtime/src/runtime.rs
stella3d: with `.map_err()` ?  I think for this one we should maybe keep it as-is, i think the flow is clearest this way.  open to suggestions on how to change it if i'm missing something

language/move-vm/runtime/src/runtime.rs
stella3d: that sounds fine - mostly I wanted to group all of these just because they're immediately repeated.  Would this make `borrow_arg` more widely useful ?

language/move-vm/integration-tests/src/tests/exec_func_effects_tests.rs
sblackshear: The Move code shouldn't return a value that fails to serialize unless there's a bug in the VM, so this would indeed be tricky without some kind of fault injection... This test looks great, thanks!

language/move-vm/runtime/src/runtime.rs
sblackshear: Yes, I think this would help! And the rename is probably a good idea because "arg" and "actual" are synonyms (e.g., like "parameter" and "formal").

sblackshear (changes requested):

language/move-vm/runtime/src/runtime.rs
stella3d: changed !

sblackshear (approved):

sblackshear:
@stella3d this looks good to go to me except for the merge conflict.

sblackshear (dismissed):

sblackshear (approved):

sblackshear (dismissed):
/land

sblackshear (dismissed):
/land

sblackshear (approved):
/land

sblackshear (approved):
/land

sblackshear (dismissed):
/land

sblackshear (approved):
/land

sblackshear (approved):
/land

sblackshear (approved):
/land

PR #32, opened 2022-01-29
[move-tools] Serde-friendly type layout generation from modules

awelc:
This PR resurrects an earlier [PR](https://github.com/diem/diem/pull/9424) by @sblackshear and at this point it mainly adapts the older implementation to the updated Move codebase with the same functionality. 

## Motivation

It's the same as before, that is for this change to be combined with serde-generate to codegen type definitions for Move structs in a variety of languages.

- Add logic for generating serde-friendly layout YAML for Move struct definitions. Only works for closed types (i.e., no unbound type parameters).
- The generation works in two modes: deep (recursively generate layout YAML for given struct and all its deps) or shallow (given struct only, leave deps as typenames).
- Exposed generation in both modes via CLI + added tests.

Yes

## Test Plan

Additional tests verifying the new layout format have been added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1715,6 +1715,7 @@ dependencies = [
  "move-binary-format",
  "move-core-types",
  "petgraph 0.5.1",
+ "serde-reflection",
  "workspace-hack",
 ]
 
@@ -3560,6 +3561,17 @@ dependencies = [
  "num-traits 0.1.43",
  "regex",
  "serde 0.8.23",
+]
+
+[[package]]
+name = "serde-reflection"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "167450ba550f903a2b35a81ba3ca387585189e2430e3df6b94b95f3bec2f26bd"
+dependencies = [
+ "once_cell",
+ "serde 1.0.130",
+ "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


awelc:
It would be great to get some additional feedback on the comments at the original PR. In particular this one

> We have AccountAddress generated already as diem core types, so we probably need a cli argument to specify where is known types package and module and use them instead of generating them, I think current transaction builder cli has the option to do so for using diem core types (includes AccountAddress) when generating stdlib functions.

From what I understand, this would allow us to avoid hard-coded constants for internal Move types, such as `AccountAddress` but it would be great to get some additional explanation on the suggestion above.

sblackshear:
> From what I understand, the last part would allow us to avoid hard-coded constants for internal Move types, such as `AccountAddress` but it would be great to get some additional explanation on the suggestion(s) above.

These comments are about the integration of the "type layout file" feature introduced by this PR with the `serdegen` code generation tool. I think integration with `serdegen` is a good goal for the future, but the type layout files are still quite useful even if they don't work with `serdegen`. For example: a wallet reads the raw bytes of a Move value of type `T` from on-chain. If the wallet also has a type layout file for `T`, it can use the file + `serde` to transform the raw bytes of the value into any tree-structured format serde can output (e.g., JSON).

Given this, I think a reasonable path forward might be:
- go ahead with this PR as is
- file a task for making `serdegen` work with type layout files with @xli's comments

Let me know if that sounds reasonable!

sblackshear (dismissed):

PR #34, opened 2022-01-30
Fixed dev_setup script to download the right version of .NET for M1 Macs

awelc:
## Motivation

This attempt to fix the problem with dev_setup.sh on M1 Mac (issue #32). The problem is that Move prover at this point needs .NET 5.x but with the current settings the script will try to download a version of .NET for the Arm64 architecture which does not exist for .NET 5.x. The solution is to force downloading the x64 version which will work just fine in emulation mode.

## Test Plan

Tested locally on an M1 MacBook Pro - the script no longer fails and prover tests run correctly.


sblackshear (approved):
Accepting + landing to unblock installation for new Move users with M1's, but copying selected prover/Boogie folks just in case.

PR #36, opened 2022-01-31
bug fix when executing command 'move sandbox run'

JerryKwan:
fix the logical bug when executing command 'move sandbox run'

## Motivation

Sometimes when executing command 'move sandbox run' in a package, it will prompt the error message such as "Unable to find script in file xxxx"
When digging more deeper of the codes, It seems like the logic in the current implementation of the command  'move sandbox run'  is not correct all the time, it just compares the filename between the compiled packages and the user inputted script name, but move-compiler has done some renaming when compiling the source codes. It would be better to compare the FileHash other than the file name to determine the correct compiled script to execute 

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/tools/move-cli/src/sandbox/commands/run.rs
+++ b/language/tools/move-cli/src/sandbox/commands/run.rs
@@ -52,9 +52,10 @@ move run` must be applied to a module inside `storage/`",
         fs::read(script_path)?
     } else {
         // TODO(tzakian): support calling scripts in transitive deps
+        let file_contents = std::fs::read_to_string(script_path)?;
         let script_opt = package
             .scripts()
-            .find(|unit| unit.source_path.file_stem() == script_path.file_stem());
+            .find(|unit| unit.unit.source_map().check(&file_contents));
         // script source file; package is already compiled so load it up
         match script_opt {
             Some(unit) => unit.unit.serialize(),


JerryKwan:
This is the solution of the issue https://github.com/diem/move/issues/35

language/tools/move-package/src/compilation/compiled_package.rs
```diff
@@ -760,7 +760,7 @@ impl CompiledPackage {
                 CompiledUnit::Module(_) => CompiledPackageLayout::CompiledModules.path(),
             };
             let path = match &compiled_unit.unit {
-                CompiledUnit::Script(named) => named.name.as_str(),
+                CompiledUnit::Script(_) => compiled_unit.source_path.file_name().unwrap().to_str().unwrap(),
```
tnowacki, line 763: There might be more than one script per file, so I don't think this works. 

Script names should be distinct from each other and from modules. Do you have a repo for this issue? Because I don't think this is quite the right fix

tnowacki: I see the repro in the issue now. Maybe @tzakian or I can take a look

tzakian: Looking at the repro, I feel as though this is more of an issue with the CLI interface for running a script than with the script naming -- I think keeping the script naming based on the actual script name instead of the file makes sense for the reason Todd mentioned.

I would propose as a different solution that the way scripts are invoked from the CLI be changed instead so that you invoke the script by the scripts name, and not by the filename of the script. e.g., instead of:

```
sandbox run sources/test__script.move
```

You would write

```
sandbox run test
```

JerryKwan: @tzakian Thanks for the suggestion
I think 
```sh
sandbox run test
```
is much better than 
```sh
sandbox run sources/test__script.move
```
But as https://diem.github.io/move/modules-and-scripts.html says, 
```raw
A script block must start with all of its use declarations, followed by any constants and (finally) the main function declaration. The main function can have any name (i.e., it need not be called main), is the only function in a script block, can have any number of arguments, and must not return a value.
```
and https://diem.github.io/move/creating-coins.html shows the using convention
```sh
move run src/scripts/test-coin.move
```
That's the reason why I send this PR.
@tzakian @tnowacki Script names (to be exactly it is full file name) should be distinct, but the main function can be identical between different scripts, and in the current version of Move, the same main function will be compiled and saved to disk with a number suffix, such as this
```text
tree .
.
├── BuildInfo.yaml
├── bytecode_modules
├── bytecode_scripts
│   ├── test_0.mv
│   └── test_1.mv
├── source_maps
│   ├── test_0.mvsm
│   └── test_1.mvsm
└── sources
    ├── Coin.move
    ├── test_script2.move
    └── test_script.move
```

tnowacki: Thats more of a compiler convention that anything else. The CLI and package system should just yell at you for duplicate script names

tnowacki: In short, the compiler is a lot more configurable and permissive than the package system or CLI. The package system and CLI should be a bit more opinionated here

JerryKwan: Agree. It's should be solved from the end of move-package and move-cli

tzakian: Not sure if you want to implement these checks or not @JerryKwan, If not I'm more than happy to implement them. Let me know!

JerryKwan: @tzakian I'd like to implement these checks. And how about we merge this PR as the solution of the current problem of move-compiler, then it will correspond to https://github.com/diem/move/blob/main/language/documentation/book/src/creating-coins.md and https://github.com/diem/move/blob/main/language/documentation/book/src/modules-and-scripts.md and reduce confusions to beginners. And we can add more functions to move-compiler and move-cli to make them more flexible.

tzakian: I'd probably prefer if we didn't merge this as-is as I feel the change is just going to be something that will then need to be changed in the future PR for the final fix. My suggestion for a near-term solution would probably be to update the docs as that will need to be done anyways, and that should hopefully remove the source of confusion.

tnowacki (changes requested):
Agreed with @tzakian, any changes here would have to be either doc related, or changing the CLI/package system to yell about duplicate scripts

Another fun possible thing would be for the CLI to say "Did you mean... ?" when invoking various commands but that is probably a considerable amount of work

language/tools/move-package/src/compilation/compiled_package.rs
JerryKwan: Updating the docs to reflect the source codes logic could be be solution for a near term, but I think there are a lot of copies of the current version of documents

JerryKwan: > Thats more of a compiler convention that anything else. The CLI and package system should just yell at you for duplicate script names

@tnowacki 
If we need to implement this function, it need to modify the codes located in move\language\move-compiler\src\expansion\translate.rs, to be more precisely we need to change the statements under function program()
```rust
pub fn program(
    compilation_env: &mut CompilationEnv,
    pre_compiled_lib: Option<&FullyCompiledProgram>,
    prog: P::Program,
) -> E::Program {
    let module_members = {
        let mut members = UniqueMap::new();
        all_module_members(
            compilation_env,
            &mut members,
            true,
            &prog.source_definitions,
        );
        all_module_members(compilation_env, &mut members, true, &prog.lib_definitions);
        if let Some(pre_compiled) = pre_compiled_lib {
            assert!(pre_compiled.parser.source_definitions.is_empty());
            all_module_members(
                compilation_env,
                &mut members,
                false,
                &pre_compiled.parser.lib_definitions,
            );
        }
        members
    };

    let mut context = Context::new(compilation_env, module_members);

    let mut module_map = UniqueMap::new();
    let mut scripts = vec![];
    let P::Program {
        source_definitions,
        lib_definitions,
    } = prog;

    context.is_source_definition = true;
    for def in source_definitions {
        definition(&mut context, &mut module_map, &mut scripts, def)
    }

    context.is_source_definition = false;
    for def in lib_definitions {
        definition(&mut context, &mut module_map, &mut scripts, def)
    }

    let mut scripts = {
        let mut collected: BTreeMap<Symbol, Vec<E::Script>> = BTreeMap::new();
        for s in scripts {
            collected
                .entry(s.function_name.value())
                .or_insert_with(Vec::new)
                .push(s)
        }
        let mut keyed: BTreeMap<Symbol, E::Script> = BTreeMap::new();
        for (n, mut ss) in collected {
            match ss.len() {
                0 => unreachable!(),
                1 => assert!(
                    keyed.insert(n, ss.pop().unwrap()).is_none(),
                    "ICE duplicate script key"
                ),
                _ => {
                    for (i, s) in ss.into_iter().enumerate() {
                        let k = format!("{}_{}", n, i);
                        assert!(
                            keyed.insert(k.into(), s).is_none(),
                            "ICE duplicate script key"
                        )
                    }
                }
            }
        }
        keyed
    };

    super::unique_modules_after_mapping::verify(context.env, &module_map);
    super::dependency_ordering::verify(context.env, &mut module_map, &mut scripts);
    E::Program {
        modules: module_map,
        scripts,
    }
}
```

JerryKwan: @tnowacki @tzakian 
The root cause of this problem is the wrong processing of source file name when loading compiled packages from the disk. The current implementation just use the file_stem() of the compiled byte code as the file name of the source_path, but `move sandbox run` compares the compiled scripts source_path with the user inputted script_path. But move-package has changed the source names when invoke save_to_disk() according the names return by move-compiler\src\expansion\translate.rs, the format is something like mainfunctionname or mainfunctionname_suffix, so there will be errors such as "Error: Unable to find script in file "sources/test_script.move""
please see the source code snippets listed below
*move/language/tools/move-package/src/compilation/compiled_package.rs/into_compiled_package()*
```rust
                let bytecode_path = Path::new(bytecode_path.as_str());
                let file_stem = bytecode_path.file_stem().unwrap();
                let bytecode_bytes = std::fs::read(&bytecode_path)?;
                let source_map = source_map_from_file(
                    &CompiledPackageLayout::SourceMaps
                        .from_sibling_path(bytecode_path)
                        .ok_or_else(|| anyhow::format_err!("Unable to find source map"))?
                        .join(file_stem)
                        .with_extension(SOURCE_MAP_EXTENSION),
                )?;
                let source_path = CompiledPackageLayout::Sources
                    .from_sibling_path(bytecode_path)
                    .ok_or_else(|| anyhow::format_err!("Unable to find source file"))?
                    .join(file_stem)
                    .with_extension(MOVE_EXTENSION);
                match CompiledScript::deserialize(&bytecode_bytes) {
                    Ok(script) => {
                        let name = FileName::from(
                            bytecode_path
                                .file_stem()
                                .unwrap()
                                .to_string_lossy()
                                .to_string(),
                        );
                        let unit = CompiledUnit::Script(NamedCompiledScript {
                            name,
                            script,
                            source_map,
                        });
                        Ok(CompiledUnitWithSource { unit, source_path })
                    }
```
*move/language/tools/move-cli/src/sandbox/commands/run.rs/run()*
```rust
       let script_opt = package
            .scripts()
            .find(|unit| unit.source_path.file_stem() == script_path.file_stem());
        // script source file; package is already compiled so load it up
        match script_opt {
            Some(unit) => unit.unit.serialize(),
            None => bail!("Unable to find script in file {:?}", script_path),
        }
```

JerryKwan: And there is an easy way to fix the error such as "Unable to find script in file"
we can change the codes in /move/language/tools/move-cli/src/sandbox/commands/run.rs
```rust
            let script_opt = package
            .scripts()
            .find(|unit| unit.source_path.file_stem() == script_path.file_stem());
        // script source file; package is already compiled so load it up
        match script_opt {
            Some(unit) => unit.unit.serialize(),
            None => bail!("Unable to find script in file {:?}", script_path),
        }
```
to codes
```rust
              // calculate file hash of the script_path
              let file_contents = std::fs::read_to_string(script_path)?;
              let file_hash = FileHash::new(&file_contents);
              println!("file_hash of {:?} is {:?}", script_path, file_hash);
              let script_opt = package
                  .scripts()
                  .find(|unit| unit.unit.source_map().check(&file_contents));
              // script source file; package is already compiled so load it up
              match script_opt {
                  Some(unit) => unit.unit.serialize(), 
                  None => bail!("Unable to find script in file {:?}", script_path),
              }
```

JerryKwan (commented):
@tnowacki @tzakian 
We can keep the logic of the current implementation of move-compiler and move-package, the only change needed to be done is changing the filter logic in /move/language/tools/move-cli/src/sandbox/commands/run.rs, using FileHash other than file name to get the correct compiled script is much more reasonable.
I have changed the codes already, would you please help to review? thank you

JerryKwan (commented):
use FileHash to filter the correct compiled package

JerryKwan:
It seems like the PR includes a merge commit, which fails the CI test, How about I re-create the PR?

tzakian:
Hey @JerryKwan you should be able to fix this up without starting a new PR by  running an interactive rebase/squashing the different commits into one single commit for this PR (see the accepted answer in this SO thread for more info: https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git).

In general we operate on a one-commit to one-PR basis here, and keep a rebase/amend/force-push cycle for PRs.

JerryKwan:
> Hey @JerryKwan you should be able to fix this up without starting a new PR by running an interactive rebase/squashing the different commits into one single commit for this PR (see the accepted answer in this SO thread for more info: https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git).
> 
> In general we operate on a one-commit to one-PR basis here, and keep a rebase/amend/force-push cycle for PRs.

Thanks @tzakian , I have rebased the commits already.

tzakian (approved):
/land

JerryKwan:
@tzakian 
Seems like it is still waiting @tnowacki to review

PR #40, opened 2022-02-04
ERC20 and ERC1155 implementations with Table

junkil-park:
This PR adds the Table module and ERC20 and ERC1155 implementations using Table.

Most of the functions in the Table module are declared to be native. They may be compiled into the Solidity-like maps by the Move-to-EVM compiler. Moreover, although it is slow, it's possible to run them on MoveVM today because they can be easily implemented using vector in Move. It can become faster in the future once Move supports an efficient map type.

As for the ERC20 contract, the differences from the original implementation (Token.move) are as follows:
- Table is used
- All data structures go to the single struct called `State` (which correspond to the contract state in Solidity)
  - State is published when the contract is deployed and the constructor is called.
  - signers are not required for adding entries to the tables
- In order to make the names compatible, the camel case is used instead of the snake case.
- events are emitted accordingly.

## Motivation

To add the Table module, and to rewrite the ERC20 contract using Table

Yes

## Test Plan

cargo test

### DIFF
--- a/language/evm/examples/sources/Table.move
+++ b/language/evm/examples/sources/Table.move
@@ -1,0 +1,35 @@
+/// This module defines Table for EVM.
+module Evm::Table {
+    native struct Table<K, V> has store;
+
+    /// Create an empty Table.
+    native public fun empty<K, V>(): Table<K, V>;
+
+    /// Acquire an immutable reference to the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    native public fun borrow<K, V>(table: &Table<K, V>, key: &K): &V;
+
+    /// Acquire a mutable reference to the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    native public fun borrow_mut<K, V>(table: &mut Table<K, V>, key: &K): &mut V;
+
+    /// Acquire a mutable reference to the value which `key` maps to.
+    /// Insert the pair (`key`, `default`) first if there is no entry for `key`.
+    public fun borrow_mut_with_default<K: copy + drop, V: drop>(table: &mut Table<K, V>, key: K, default: V): &mut V {
+        if (!contains(table, &key)) {
+            insert(table, copy key, default)
+        };
+        borrow_mut(table, &key)
+    }
+
+    /// Remove from `table` and return the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    native public fun remove<K, V>(table: &mut Table<K, V>, key: &K): (K, V);
+
+    /// Returns true iff `table` contains an entry for `key`.
+    native public fun contains<K, V>(table: &Table<K, V>, key: &K): bool;
+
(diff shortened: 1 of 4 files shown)


language/evm/examples/sources/Table.move
```diff
@@ -0,0 +1,35 @@
+/// This module defines Table for EVM.
+module Evm::Table {
+    native struct Table<K, V> has store, copy, drop;
+
```
wrwg, line 4: I guess a table should not have copy and perhaps also not drop, because it is a potentially very large data structure.

language/evm/examples/sources/Table.move
```diff
@@ -0,0 +1,35 @@
...
+    /// Create an empty Table.
+    native public fun empty<K, V>(): Table<K, V>;
+
+    /// Acquire an immutable reference to the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    native public fun borrow<K, V>(table: &Table<K, V>, key: &K): &V;
+
+    /// Acquire a mutable reference to the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    native public fun borrow_mut<K, V>(table: &mut Table<K, V>, key: &K): &mut V;
```
wrwg, line 14: This may in fact need to extend the prover's borrow analysis and perhaps also Move compiler. The borrow_mut on vectors is hard wired in those, as it creates a borrow edge labelled by the vector index; something similar is needed here. Not that this is a blocker for this, but just need to keep this additional complexity in mind.

language/evm/examples/sources/Table.move
junkil-park: Done. Good point!

language/evm/examples/sources/Table.move
junkil-park: OK. I will check this with Todd later.

wrwg (dismissed):

wrwg (commented):
Oh there is a new file for 1155? Let me check first before you land.

wrwg:
> Oh there is a new file for 1155? Let me check first before you land.

Never mind, looks good.

language/evm/examples/sources/Table.move
```diff
@@ -0,0 +1,35 @@
+/// This module defines Table for EVM.
+module Evm::Table {
```
tnowacki, line 2: Maybe there was some confusion around naming things, but these are just maps, not tables. Tables are a new scalable storage type that will sit alongside the current resource system. 

Mind just renaming this to `Map` :D 

Looks good otherwise though!

wrwg: This is supposed to be scalable storage type thing. For example, they aren't enumerable. Also the scalable storage thing we need for EVM Move contracts. Perhaps let's review this API from this viewpoint again? Nothing is set in stone right now regards how it is defined.

PR #44, opened 2022-02-07
[utils] add logic for serializing Move value with both type and field…

sblackshear:
… names

This is useful for viewing Move values as tree-structured data in formats that are likely to be consumed by clients (e.g., json).

### DIFF
--- a/language/move-core/types/src/language_storage.rs
+++ b/language/move-core/types/src/language_storage.rs
@@ -18,13 +18,21 @@ pub const CORE_CODE_ADDRESS: AccountAddress = AccountAddress::ONE;
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Hash, Eq, Clone, PartialOrd, Ord)]
 pub enum TypeTag {
+    #[serde(rename = "bool")]
     Bool,
+    #[serde(rename = "u8")]
     U8,
+    #[serde(rename = "u64")]
     U64,
+    #[serde(rename = "u128")]
     U128,
+    #[serde(rename = "address")]
     Address,
+    #[serde(rename = "signer")]
     Signer,
+    #[serde(rename = "vector")]
     Vector(Box<TypeTag>),
+    #[serde(rename = "struct")]
     Struct(StructTag),
 }
 
@@ -34,6 +42,7 @@ pub struct StructTag {
     pub module: Identifier,
     pub name: Identifier,
     // TODO: rename to "type_args" (or better "ty_args"?)
+    #[serde(rename = "type_args")]
     pub type_params: Vec<TypeTag>,
 }
 
--- a/language/move-core/types/src/unit_tests/value_test.rs
+++ b/language/move-core/types/src/unit_tests/value_test.rs
@@ -1,0 +1,103 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    account_address::AccountAddress,
+    ident_str,
+    identifier::Identifier,
+    language_storage::{StructTag, TypeTag},
+    value::{MoveStruct, MoveValue},
+};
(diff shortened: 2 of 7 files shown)


language/move-core/types/src/unit_tests/value_test.rs
```diff
@@ -0,0 +1,103 @@
...
+        serde_json::to_value(&fielded_value).unwrap(),
+        json!({ "f": 7, "g": true })
+    );
+
+    let typed_value = MoveStruct::with_types(struct_type.clone(), field_values.clone());
+    assert_eq!(
+        serde_json::to_value(&typed_value).unwrap(),
+        json!({
+                "fields": { "f": 7, "g": true },
+                "type": "0x0::MyModule::MyStruct"
```
tzakian, line 44: I'm wondering if this type should be even further deconstructed possibly. e.g.,

```json
"type": {
    "address": "0x0",
    "module_name": "MyModule",
    "type_name": "MyStruct"
}
```

In the monomorphic case I don't think there is much of a difference, but in the generic case it may make it easier to distinguish/find multiple instantiations of the same generic type:

```json
[{
  "fields": { ... },
  "type": {
     "address": "0x0",
     "module_name": "MyModule",
     "type_name": "GenericStruct",
     "instantiation": [ "u8" ]
   }
},
{
  "fields": { ... },
  "type": {
     "address": "0x0",
     "module_name": "MyModule",
     "type_name": "GenericStruct",
     "instantiation": [ "u64" ]
   }
}]
```

tzakian (approved):
This looks good to me. I'm fine with this going in, but my main question is about the representation of types and whether those should be JSON objects themselves

language/move-core/types/src/unit_tests/value_test.rs
sblackshear: Very good question. I actually did it this way initially, but ended up changing my mind and going with the current approach. My reasoning:
- I think clients of the code will most commonly want to use the fully qualified type directly rather than breaking it into its components. E.g.: map the type to its Python or JS equivalent + create a Python or JS object for the value, index the struct value by its type, pipe it into a query like "show me other values of this type", copy the type and paste it into a block explorer. or just view the type in a human-readable format.
- This way ends up being quite a bit more human-readable, especially considering that every struct in a (potentially deep) nesting has an attached type. If you use "type as object representation", I think any nested struct quickly becomes quite difficult to grok.

With that said, I think we should probably support multiple ways of printing the types just as we support multiple ways of printing struct values. Some refactoring of value.rs is probably needed to support this, the current approach is already becoming a bit unwieldy...

sblackshear:
@tzakian for re-stamp plz!

language/move-core/types/src/unit_tests/value_test.rs
tzakian: Good points, and agreed that this is becoming a bit unwieldy. Eventually I think it would be best to have a config where these types of things can be set ("packed" vs "unpacked" type representations in the output).

tzakian (approved):

language/move-core/types/src/language_storage.rs
```diff
@@ -18,13 +18,21 @@ pub const CORE_CODE_ADDRESS: AccountAddress = AccountAddress::ONE;
...
     U64,
+    #[serde(rename = "u128")]
     U128,
+    #[serde(rename = "address")]
     Address,
+    #[serde(rename = "signer")]
     Signer,
+    #[serde(rename = "vector")]
     Vector(Box<TypeTag>),
+    #[serde(rename = "struct")]
```
areshand, line 35: Hi @sblackshear  could I change this serde rename to "Struct" as well as other rename to capital case? Since "struct" is a reserved keyword in rust, the code deserialized back to RUST won't compile.

PR #50, opened 2022-02-12
[cleanup] return Result instead of panic-ing in TypeLayoutBuilder

sblackshear:
No need to panic here, just makes life harder for a caller that made a mistake.


wrwg (approved):

sblackshear:
Had to fix a lint error, re-stamp requested!

wrwg (approved):

PR #52, opened 2022-02-14
[move-to-yul] Memory model

wrwg:
This implements the basic memory model for Move.

- Locals and references to them. Locals which are borrowed need to be evaded to memory so one can take a reference of them.
- Structs with Pack and BorrowField.
- ReadRef and WriteRef
- Yul functions to allocate and free memory, and to load and store values of different sizes.

This does not yet deal with resources in storage (borrow_global et. al); however, the logic for reading storage as linear memory is there.

This PR also changes the baseline output to include the optimized Yul as generated by solc. To this end, as the dispatcher isn't there yet. we needed to insert some dummy calls for all `#[callable]` as otherwise the optimizer will remove unused functions. The optimizations on Yul level are quite impressive!

As there are no semantic tests yet, I'm sure there are multiple errors in this code. Those will be fixed once testing infra is up.

## Motivation

EVM

Yes

## Test Plan

Baseline


rahxephon89 (approved):

language/evm/move-to-yul/src/yul_functions.rs
```diff
@@ -106,12 +120,259 @@ fn make_yule_def(name: &str, body: &str) -> String {
...
+// Loads u8 from memory offset.
+MemoryLoadU8: "(offs) -> val {
+  val := and(mload(offs), ${MAX_U8})
+}",
+
+// Loads u8 from storage offset.
+StorageLoadU8: "(offs) -> val {
+  let word_offs, bit_offs := $ToWordOffs(offs)
+  let key := $StorageKey(${CONTINUOUS_STORAGE_TYPE}, word_offs)
+  val := and(shr(sload(key), bit_offs), ${MAX_U8})
```
emmazzz, line 203: Are we using little endian within the word here, where smaller bit offset stores less significant bit? So bit offset 0 is actually the least significant bit of the 256 bits. This seems a bit counterintuitive. If we want to use big endian instead we should `shr` by `(256 - bit_offs - 8)`.

Same confusion for the other storage load and store functions.

wrwg: This was written under the wrong assumption of little endian, and is fixed in PR #79

PR #53, opened 2022-02-15
[move-compiler] Allow implicit copies on all copyable types

tnowacki:
- Allow implicit copy on any value whose type has the copy ability
- Switch last usage of every variable to a move
- Refuse/error to infer copy/move if the local is borrowed during its last usage

## Motivation

- It has been a source of confusion for a while
- The original motivation of only allowing primitives was to save gas. But it doesn't really help with that in a general setting. Sure, it helps in this one narrow case, but there are so many potential gas issues that I don't think accidental duplication is at the top of the list. Hopefully this new assumption isn't wrong

## Test Plan

- new test

### DIFF
--- a/language/evm/move-to-yul/tests/Locals.exp
+++ b/language/evm/move-to-yul/tests/Locals.exp
@@ -55,9 +55,9 @@ object "A2_M" {
                 $t13 := $AddU64($t11, $t12)
                 // write_ref($t9, $t13)
                 $StoreU64($t9, $t13)
-                // $t14 := copy($t0)
+                // $t14 := move($t0)
                 $t14 := mload($locals)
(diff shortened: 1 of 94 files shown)


wrwg (approved):

sblackshear (approved):

sblackshear (approved):

wrwg:
There seems to be non-determinism in the yul baseline tests. Looking at a fix.

wrwg:
> There seems to be non-determinism in the yul baseline tests. Looking at a fix.

Once #86 lands and after rebase, this should be gone.

sblackshear (approved):
/land

wrwg (approved):

PR #56, opened 2022-02-15
[evm] initial implementation of evm execution utils

vgao1996:
This introduces a crate `evm-exec-utils`, which contains some wrappers around compiling solidity code (for testing) and executing EVM bytecode using [SputnikVM](https://github.com/rust-blockchain/evm). 

Right now, this has limited functionalities and some of the parameters might not having been properly set up. We'll explore and fix them over time.


wrwg (dismissed):
Looks good to me! I think we should move forward with this and refine it as we go.

PR #57, opened 2022-02-15
[evm] add support for U256

emmazzz:
## Motivation

This PR adds support for u256 in Move to Yul translation:
- Added a U256 module defining a native struct representing u256 and other arithmetic methods
- Added U256 as a constant type in stackless bytecode and CastU256 as an operation
- Added a step of stackless bytecode transformation for evm
- Added corresponding yul functions
- Fixed a little issue with liveness analysis so we get rid of unnecessary loads as well (before we only remove assignments)


language/evm/move-to-yul/src/evm_transformation.rs
```diff
@@ -0,0 +1,121 @@
...
+            {
+                if crate::generator::is_evm_arith_module(
+                    &func_env.module_env.env.get_module(*module_id),
+                ) {
+                    let fun_name = &*func_env
+                        .module_env
+                        .env
+                        .get_module(*module_id)
+                        .get_function(*fun_id)
+                        .get_simple_name_string();
```
wrwg, line 53: Should this be better the qualified name you check for?

wrwg (approved):

language/evm/stdlib/sources/U256.move
```diff
@@ -0,0 +1,18 @@
+#[evm_arith]
+module 0x1::U256 {
+    native struct U256 has copy, drop, store;
+    native public fun u256_from_words(lo: u128, hi: u128): U256;
+    native public fun add(x: U256, y: U256): U256;
```
wrwg, line 5: I'm wondering whether it would be possible to extend the frontend to allow for `x + y` etc., and be able to use u256 (small case) without import (like vector),  Then at the moment where the bytecode is generated we insert the call to the native function. The current way how we represent this doesn't seem to create a good UX compared to Solidity. @tnowacki @vgao1996

language/evm/move-to-yul/src/evm_transformation.rs
emmazzz: Qualified name meaning something like "0x1::U256::add"? Since earlier we already checked on line 45 the module identified by module_id is an evm_arith module, I think it's sufficient to check simple name?

wrwg: Ah ok.

wrwg (dismissed):

wrwg (dismissed):

language/evm/move-to-yul/tests/U256Arith.exp
```diff
@@ -0,0 +1,332 @@
...
+                r := add(shl(hi, 128), lo)
+            }
+        }
+    }
+}
+
+
+!! Errors compiling Yul
+
+Warning: Yul is still experimental. Please use the output with care.
```
wrwg, line 327: Sorry for not noticing previously, but there is a problem with these source annotations.

wrwg: Not sure whether this is the right solution. It could be a regular comment which just ignores the content? Perhaps one can see from the optimized Yul dump.

PR #60, opened 2022-02-17
[language] Remove outdated tutorial

666lcz:
## Motivation

This is an old tutorial (predates the Move package system!) that has been subsumed by https://github.com/diem/move/tree/main/language/documentation/tutorial

Yes

## Test Plan

Make sure clicking the links goes to the right page on VS Code Markdown Viewer


language/documentation/book/src/overview.md
```diff
@@ -27,8 +27,8 @@ Move is a next generation language for secure, sandboxed, and formally verified
     icon="img/placeholder.svg"
     iconDark="img/placeholder-dark.svg"
     overlay="Play with Move directly as you create coins with the language"
-    title="First Tutorial: Creating Coins"
-    to="/docs/move/move-tutorial-creating-coins"
+    title="Move Tutorial"
+    to="https://github.com/diem/move/tree/main/language/documentation/tutorial"
```
666lcz, line 31: I don't think the tutorial lives in the `/docs/`, therefore using an absolute path here. But I am not sure how this will get rendered: https://github.com/666lcz/move/blob/02-16-update-tutorial/language/documentation/book/src/overview.md

tzakian: If you want to preview it locally you'll need `mdbook` installed. After that you can preview local changes by running `mdbook serve --open` and that should open a browser with the current changes. 

In order to update the website you'll need to rebuild the mdbook (I usually run `mdbook build` for this and make sure the generated files are in the correct location/submitted to the `gh-pages` branch of this repo). More information on `mdbook` can be found here: https://rust-lang.github.io/mdBook/guide/creating.html

666lcz:
Thanks @tzakian. According to the [mdbook documentation](https://rust-lang.github.io/mdBook/guide/creating.html#summarymd), the link for a chapter in `SUMMARY.md` must be a relative path within the `book/src` directory, therefore to replace the old tutorial with the new one(which resides outside the `book/src` directory, we need to choose one of the following:

1. Keep the old tutorial file but replace the content with a link to the new tutorial;
2. Move the new tutorial from `language/documentation/tutorial` to `language/documentation/book/src`. (This is not ideal because the original tutorial folder contains various code folders).
3. Duplicate content in two places(easy to introduce consistency).

I updated the revision with the first option and verified the change locally by using `md serve --open`. @sblackshear let me know if you have any concern.

sblackshear (approved):
/land

PR #64, opened 2022-02-17
[move-compiler] Allow different address mappings per file

tnowacki:
- Updated source language compiler to allow multiple address mappings
- A different mapping can be specified per file
- Feature not yet utilized

## Motivation

- Will allow the package system to use source files 

## Test Plan

- ran em


sblackshear (dismissed):

tzakian (dismissed):
Looks good to me. Thanks for doing this Todd!

PR #68, opened 2022-02-18
[move-to-yul] Add missing implementation of MoveFrom

wrwg:
This extends the persistent memory implementation by handling MoveFrom.  At this point, we covered all opcodes! However, we are still missing:

- A concept for implementing native functions.
- Vectors as one use case of native functions, including their memory layout.
- Testing that all this basic stuff works

The PR also refactors some of the MoveTo code, and uses a more efficient struct layout where fields which fit on word boundaries are allocated first. Some bugs discovered from reading the generated Yul are also fixed. (There are certainly many more.)

## Motivation

EVM

Yes

## Test Plan

New baseline test for MoveFrom


rahxephon89 (approved):

language/evm/move-to-yul/src/yul_functions.rs
```diff
@@ -152,7 +154,8 @@ AbortBuiltin: "() {
 //   arena style.
 Malloc: "(size) -> offs {
     offs := mload(${MEM_SIZE_LOC})
-    mstore(${MEM_SIZE_LOC}, add(offs, size))
+    // pad to word size
+    mstore(${MEM_SIZE_LOC}, add(offs, shr(add(size, 31), 5)))
```
emmazzz, line 158: Here I suppose we are padding `size` to a multiple of 32. If so, did we miss a `shl` by 5? Should this line be the following instead:
```
mstore(${MEM_SIZE_LOC}, add(offs, shl(shr(add(size, 31), 5),5)))
```

language/evm/move-to-yul/src/generator.rs
```diff
@@ -1238,102 +1253,266 @@ impl Generator {
...
+                );
+                // Load the pointer to the linked memory.
+                emitln!(
+                    ctx.writer,
+                    "let $linked_src := mload(add({}, {}))",
+                    src,
+                    byte_offs
+                );
+                // Recursively move.
+                let field_struct_id = ty.get_struct_id(ctx.env).expect("struct");
```
emmazzz, line 1307: The field can be of a primitive type, not necessarily a struct, right?

language/evm/move-to-yul/src/yul_functions.rs
wrwg: Fixed in PR #79  (also the wrong order of operands to shift functions)

language/evm/move-to-yul/src/generator.rs
wrwg: ... or to vectors. When vectors come, this code will become a bit more complex.

PR #75, opened 2022-02-22
control flow graph to structured control flow

gegrid:
## Motivation

(Write your motivation for proposed changes here.)

(Write your answer here.)

## Test Plan

## Related PRs


language/evm/move-to-yul/tests/ControlStructures.exp.apply-cfg-to-scf
```diff
@@ -0,0 +1,370 @@
...
+}
+
+
+!! Errors compiling Yul
+
+Warning: Yul is still experimental. Please use the output with care.
+Error: Variable not found or variable not lvalue.
+   --> <stdin>:215:31:
+    |
+215 |                     case 0  { $block := 3 }
```
wrwg, line 339: Should we fix these errors?

language/evm/move-to-yul/tests/ControlStructures.exp.apply-cfg-to-scf
```diff
@@ -0,0 +1,370 @@
...
+                    // $t1 := 0
+                    $t1 := 0
+                    // $t2 := >($t0, $t1)
+                    $t2 := $Gt(x, $t1)
+                    // if ($t2) goto L0 else goto L2
+                    switch $t2
+                    case 0  { $block := 3 }
+                    default { $block := 2 }
+                }
+                {
```
wrwg, line 256: This is somehow wrong. The Abort block is never reached. Applies to other examples as well.

language/evm/move-to-yul/src/experiments.rs
```diff
@@ -24,4 +24,8 @@ impl Experiment {
     /// during tests this is off.
     /// Retention: permanent
     pub const CAPTURE_SOURCE_INFO: &'static str = "capture-source-info";
+    /// Transform control flow graph to structured control flow.
```
wrwg, line 27: Nit: please add a newline before new declaration.

language/evm/move-to-yul/src/functions.rs
```diff
@@ -105,30 +106,119 @@ impl<'a> FunctionGenerator<'a> {
...
 
-            // Emit state machine to represent control flow.
-            // TODO: Eliminate the need for this, see also
-            //    https://medium.com/leaningtech/solving-the-structured-control-flow-problem-once-and-for-all-5123117b1ee2
-            if cfg.successors(entry_bb).iter().all(|b| cfg.is_dummmy(*b)) {
-                // In this trivial case, we have only one block and can omit the state machine
-                if let BlockContent::Basic { lower, upper } = cfg.content(entry_bb) {
-                    for offs in *lower..*upper + 1 {
-                        self.bytecode(ctx, fun_id, target, &label_map, &code[offs as usize], false);
+            if !ctx.options.apply_cfg_to_scf() {
```
wrwg, line 111: This code has become too large. Can we factor the blocks in the if/else here out into a helper functions?.

language/evm/move-to-yul/src/functions.rs
```diff
@@ -105,30 +106,119 @@ impl<'a> FunctionGenerator<'a> {
...
+                        if let Some(one_loop) = loop_map.get(&blk_id) {
+                            emitln!(ctx.writer, "let $block := {}", one_loop.loop_header);
+                            emit!(ctx.writer, "for {} true {} ");
+                            ctx.emit_block(|| {
+                                if let BlockContent::Basic { lower, upper } =
+                                    cfg.content(one_loop.loop_header)
+                                {
+                                    ctx.emit_block(|| {
+                                        for offs in *lower..*upper + 1 {
+                                            self.bytecode(
```
wrwg, line 174: Should make a helper for generating code from lower...upper.

language/move-prover/bytecode/src/stackifier_topological_sort.rs
```diff
@@ -0,0 +1,168 @@
...
+    top: HashMap<BlockId, StackifierDependency>,
+    // complete topology will not change while popping
+    complete_top: HashMap<BlockId, StackifierDependency>,
+}
+
+impl StackifierDependency {
+    pub fn new() -> StackifierDependency {
+        StackifierDependency {
+            num_prec: 0,
+            succ_set: HashSet::new(),
```
wrwg, line 26: Could using a HashSet create some non-deterministic output? We typically use BTreeSet and BTreeMap. HashSet may also work if we are sure it does not influence the order in output.

language/move-prover/bytecode/src/stackifier_topological_sort.rs
```diff
@@ -0,0 +1,168 @@
...
+            .cloned()
+        {
+            nested_blocks.push(next_succ);
+        } else {
+            return None;
+        }
+        Some(nested_blocks)
+    }
+
+    // priority succ is a block whose only precessor is last block
```
wrwg, line 152: nit: make this a doc comment and the text a sentence.

language/move-prover/bytecode/src/stackless_control_flow_graph.rs
```diff
@@ -16,10 +17,10 @@ type Map<K, V> = BTreeMap<K, V>;
 type Set<V> = BTreeSet<V>;
 pub type BlockId = CodeOffset;
 
-#[derive(Debug)]
-struct Block {
-    successors: Vec<BlockId>,
-    content: BlockContent,
+#[derive(Clone, Debug)]
+pub struct Block {
```
wrwg, line 21: Can we make all those things `pub(crate)` to avoid exporting details out of this crate?

language/move-prover/bytecode/src/stackless_structured_control_flow.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::stackifier_topological_sort::StackifierTopologicalSort;
+use crate::stackless_control_flow_graph::StacklessControlFlowGraph;
+
+pub struct StacklessStructuredControlFlow {
+    pub top_sort: StackifierTopologicalSort,
```
wrwg, line 8: I think this representation of the CFG is not sufficient. You somehow need to encode conditionals. Because you aren't doing this, also the generated output doesn't deal correctly with if-then-else.

wrwg (changes requested):

language/move-prover/bytecode/src/stackless_structured_control_flow.rs
gegrid: @wrwg Thanks for the review, after the last commit, as shown in "ControlStructures.exp.apply-cfg-to-scf", the YUL codes are still using the `switch`, and the YUL errors are resolved now.
```
!! Succeeded compiling Yul
```

I might need some help to emit `do-while` and `if-else` instead, can we have a quick chat when you are free, maybe tomorrow?

language/evm/move-to-yul/tests/ControlStructures.exp.apply-cfg-to-scf
```diff
@@ -0,0 +1,309 @@
...
+                        case 3 {
+                            // label L2
+                            // $t4 := 0
+                            $t4 := 0
+                            // $t5 := ==($t0, $t4)
+                            $t5 := $Eq(x, $t4)
+                            // if ($t5) goto L4 else goto L3
+                            switch $t5
+                            case 0  { $block4 := 4 }
+                            default { $block4 := 5 }
```
wrwg, line 156: This block does not exist in this switch.

language/evm/move-to-yul/tests/ControlStructures.exp.apply-cfg-to-scf
```diff
@@ -0,0 +1,309 @@
...
+                let $block := 4
+                switch $block
+                case 4 {
+                    // $t2 := 0
+                    $t2 := 0
+                    // $t3 := >($t0, $t2)
+                    $t3 := $Gt(x, $t2)
+                    // if ($t3) goto L0 else goto L2
+                    switch $t3
+                    case 0  { $block := 3 }
```
wrwg, line 189: I think the general pattern to reduce this to an if-then-else is as follows. Given (after topsort):

```
L1:
    B1
    if c goto L2 else L3
L2:
    B2
    goto L4
L3:
    B3        
L4:
```

... becomes:

```
B1
switch c
case 0 { B3 }
default { B2 }
```

So one would need to identify these patterns n the code.

IMO one would best do this by introducing a temporary SCG data structures which represents if-then-else and possibly also loops.

wrwg: > @wrwg Thanks for the comment! I was a bit stuck and thus checked out how Cheer implemented it, I found that it had more granular "tokens", or "bytecodes" in our repo iiuc, https://github.com/leaningtech/cheerp-compiler/blob/30a05221c34f418374dd8b8149d63f0917477d34/llvm/lib/CheerpWriter/CheerpWriter.cpp#L5009-L5028
> 
> so I am wondering do we need a separate set of bytecodes other than those in "stackless_bytecodes.rs", so that we can generate if-else and while based on the more granular ones instead?

I'm not sure what you mean with finer granularity. We have a rather regular representation of bytecode and we certainly won't change any of this. The code you are pointing to seems unrelated to the problem at hand, it is just a compiler from some IR to a text form.

> 
> > a temporary SCG data structures which represents if-then-else and possibly also loops.
> 
> can you elaborate a bit how SCG here would help detect the pattern?

I'm not talking about how to detect the pattern, but how to represent the result. However, thinking about the result may  guide you also how to do this:

```
enum SCG {
   Block{ start_offs: usize, end_offs: usize } // slice of code without any branches
   If{ cond: TempIndex, if_true: Box<SCG>, if_false: Box<SCG> }
   Loop{ body: SCG } // perhaps need a new opcode for continue and break
```


> 
> Thanks a lot!

gegrid: @wrwg thanks for the explanation!

> The code you are pointing to seems unrelated to the problem at hand, it is just a compiler from some IR to a text form.

I was thinking of generating text forms with blocks, now the text forms are generated based on the related bytecode, more specifically "branch" for control flows. Contrary to this, in the code pointer, it has a notion of "token", which has different tokens for "if", "while" etc.  With the "branch" we can directly generate the switch style of text codes while to generate if-else, I guessed we would need some similar notion of "tokens" there?

>  how to represent the result

I see, this makes sense!

wrwg: > @wrwg thanks for the explanation!
> 
> > The code you are pointing to seems unrelated to the problem at hand, it is just a compiler from some IR to a text form.
> 
> I was thinking of generating text forms with blocks, now the text forms are generated based on the related bytecode, more specifically "branch" for control flows. 

That is what I mean with this SCG data structure. You generate this one, then change the code in generator.rs to generate Yul from the SCG.

Contrary to this, in the code pointer, it has a notion of "token", which has different tokens for "if", "while" etc. With the "branch" we can directly generate the switch style of text codes while to generate if-else, I guessed we would need some similar notion of "tokens" there?
> 
> > how to represent the result
> 
> I see, this makes sense!

gegrid:
the if-else logic is now emitted, per Wolfgang's suggestion, will leave the handling of loops with new bytecodes for now.

language/evm/move-to-yul/src/functions.rs
```diff
@@ -107,48 +108,260 @@ impl<'a> FunctionGenerator<'a> {
...
+            } => {
+                emitln!(ctx.writer, "switch $t{} ", cond);
+                emit!(ctx.writer, "case 0 ");
+                if let StacklessSCG::BasicBlock {
+                    start_offset,
+                    end_offset,
+                } = **if_false
+                {
+                    ctx.emit_block(|| {
+                        for offs in start_offset..end_offset + 1 {
```
wrwg, line 273: Something not right here. You are already emitting code for a BasicBlock on line 242. Why do you need to do it again here? And as said, what happens if `if_true` and `if_false` are not basic blocks? This should be a recursive algorithm here which handles the variants of StacklessSCG exactly once.

language/evm/move-to-yul/src/functions.rs
```diff
@@ -107,48 +108,260 @@ impl<'a> FunctionGenerator<'a> {
...
+                                );
+                            }
+                        })
                     }
+                }
+            })
+        }
+    }
+
+    fn emit_scf_from_cfg(
```
wrwg, line 179: As we discussed offline, the construction of the SCG really doesn't belong in this module. It should fully go into the module which defines this data type.

language/evm/move-to-yul/src/functions.rs
```diff
@@ -107,48 +108,260 @@ impl<'a> FunctionGenerator<'a> {
...
+        visited_blocks: &mut BTreeSet<BlockId>,
+        code: &[Bytecode],
+    ) {
+        let label_map = Self::compute_label_map(cfg, code);
+        let get_block = |l| label_map.get(l).expect("label has corresponding block");
+        if let BlockContent::Basic { lower, upper } = cfg.content(blk_id) {
+            let mut start = *lower;
+            for offs in *lower..*upper + 1 {
+                match &code[offs as usize] {
+                    Bytecode::Branch(_, if_t, if_f, cond) => {
```
wrwg, line 337: The branch is always at the end of the block -- this is the standard definition of basic blocks.

That's why we also do not actually need break and continue bytecodes. The SCG data structure simply could have a marker at the end of each block whether it ends with a continue or break.

language/evm/move-to-yul/src/functions.rs
```diff
@@ -283,19 +497,23 @@ impl<'a> FunctionGenerator<'a> {
...
-                emitln!(
-                    ctx.writer,
-                    "switch {}\n\
-                     case 0  {{ $block := {} }}\n\
-                     default {{ $block := {} }}",
-                    local(cond),
-                    get_block(if_f),
-                    get_block(if_t),
-                )
+                if !skip_block {
```
wrwg, line 506: Since a branch is always at the end of a basic block, your algorithm can simply construct SCG::BasicBlock(start..end) which does not contain the branch anymore (let it 'end" just before the branch).

language/evm/move-to-yul/src/functions.rs
```diff
@@ -107,48 +108,260 @@ impl<'a> FunctionGenerator<'a> {
...
+                        start = offs;
+
+                        let if_block_id = get_block(if_t);
+                        let else_block_id = get_block(if_f);
+                        let if_else_scg = StacklessSCG::IfBlock {
+                            cond: *cond,
+                            if_true: Box::new(StacklessSCG::new(*if_block_id, &cfg)),
+                            if_false: Box::new(StacklessSCG::new(*else_block_id, &cfg)),
+                        };
+                        visited_blocks.insert(*if_block_id);
```
wrwg, line 351: I'm not sure whether this is correct. You have not fully processed if/else blocks, which can contain themselves ifs and loops. Marking them as done seems to ignore them for further processing, AFAICT. This algorithm, I guess, needs to be somehow recursive.

language/evm/move-to-yul/src/functions.rs
gegrid: to make sure that I followed, "blocks" were assigned on "Branch" and "Jump", so to do it right, I want to only skip it on "Jump"? Thanks!

language/evm/move-to-yul/src/functions.rs
gegrid: IIUC, the insertion is to avoid pushing duplicates to scg_vec.
Re nested if-else, I followed your comment and counted on the processing of the scg_vec to recursively emit the inner if-else.

language/evm/move-to-yul/src/functions.rs
gegrid: sry that I am not following here, do you mean that here I should try to match other bytecodes like Assign rather than Branch?

> a marker at the end of each block whether it ends with a continue or break
is this the same idea as we chatted last time, to computing a map from label to "block position" and determine if we put a continue or break there?

Thanks for the patience!

language/evm/move-to-yul/src/functions.rs
wrwg: Each block either ends with a branch, jump, or return. Those instructions cannot appear anywhere else than at the end of a block.

language/evm/move-to-yul/src/functions.rs
wrwg: > sry that I am not following here, do you mean that here I should try to match other bytecodes like Assign rather than Branch?

I mean that you don't need to loop from lower to upper, you just look at the last instruction in the block.

> 
> > a marker at the end of each block whether it ends with a continue or break
> 
> is this the same idea as we chatted last time, to computing a map from label to "block position" and determine if we put a continue or break there?

I mean we do not need new instructions, We can rather represent break and continue like in `SCG::Block(start, end, block_kind)` where `eum BlockKind { Fallthrough, Break, Continue} ` or something like this.

> 
> Thanks for the patience!

gegrid:
Hi @wrwg I followed the idea and was able to generate "block_kind" for each block depending on its trailing goto label, but looking at the emitted codes, it's still not about right.

One issue I saw was, now the loop_header is still a basic, which is not able to handle either ifBlock or forBlock, the same thing for the if_t and if_f of IfBlock, I will need to think of ways to recursively generate the inner Block there.

Publishing the latest version now and will try to fix it, any wisdom is more than welcome.

PR #79, opened 2022-02-24
[move-to-yul] Fixes to memory primitives & passing basic reference and struct tests

wrwg:
This turns the code examples in `tests/Structs.move` and `tests/Locals.move` into tests, and fixes bugs on the way in the handling of references and structs. At this point, we have basic things like packing and unpacking structs, and working with references to locals and to fields working.

There where two major bugs in the memory primitives: (a) calling `shl` and `shr` with wrong parameter order (b) wrongly assuming low endian architecture for packing bytes in words. Unfortunately, the big endian architecture makes things a bit more tricky. So most of the primitives around LoadUNN and StoreUNN are rewritten. Notice this does not yet test storage, only memory.

## Motivation

EVM

Yes

## Test Plan

New tests


emmazzz (approved):
LGTM

PR #83, opened 2022-02-28
[move-to-yul] Semantic storage tests.

wrwg:
This PR makes the tests in `Resource.move` (those related to working with resources in storage) semantic tests and fixes issues such that they pass.

- There was an unnecessary double indirection for structs: if they are borrowed, they had been evaded to memory, like a primitive. However, as structs are already in memory, this is unnecessary (and was also buggy).
- There were a few smaller bugs around the implementation of resources in storage.
- The counter for linked storage has been moved from memory to storage where it belongs (it needs to be persistent, indeed).
- The `@use-src` directive syntax for multiple sources is no finally understood and fixed (comma separated list)
- The dumping of optimized Yul to the baseline has been removed. If we want to look it it, we can do differently.
- For debugging, I played with the "tracing" option of the Evm crate.  One can now set the environment variable EVM_STEP_LISTENER to trace the execution of every single EVM instruction. However, the output isn't very useful until now... Left the code anyway.

## Motivation

EVM

Yes

## Test Plan

New tests


language/evm/move-to-yul/src/functions.rs
```diff
@@ -149,15 +149,20 @@ impl<'a> FunctionGenerator<'a> {
...
+    /// to them.
+    fn collect_borrowed_locals(&mut self, ctx: &Context, target: &FunctionTarget) {
         let mut mem_pos = 0;
         for bc in &target.data.code {
             if let Bytecode::Call(_, _, Operation::BorrowLoc, srcs, _) = bc {
-                if let Entry::Vacant(e) = self.borrowed_locals.entry(srcs[0]) {
-                    e.insert(mem_pos);
-                    mem_pos += 1
+                let ty = target.get_local_type(srcs[0]);
+                if !ctx.type_allocates_memory(ty) {
```
emmazzz, line 161: Good catch!

language/evm/move-to-yul/src/functions.rs
```diff
@@ -901,7 +943,7 @@ impl<'a> FunctionGenerator<'a> {
             // and if it is set, abort. Otherwise set this bit.
             let exists_call = self.parent.call_builtin_str(
                 ctx,
-                YulFunction::StorageLoadU8,
+                YulFunction::AlignedStorageLoad,
```
emmazzz, line 946: We need to use this instead of LoadU8 because the RESOURCE_EXISTS_FLAG_SIZE is 32 bytes?

language/evm/move-to-yul/src/functions.rs
```diff
@@ -645,31 +650,38 @@ impl<'a> FunctionGenerator<'a> {
...
+            // A primitive which has been evaded to memory,
+            // Need to adjust the offset for the local by (32 - size) to account for big endian.
+            // We need to point to the actual highest byte of the value.
+            let offs = (*self
+                .borrowed_locals
+                .get(&srcs[0])
+                .expect("local evaded to memory")
+                * yul_functions::WORD_SIZE)
+                + 32
+                - ctx.type_size(ty);
```
emmazzz, line 667: Good catch!

emmazzz (dismissed):
LGTM!

language/evm/move-to-yul/src/functions.rs
wrwg: Yes, its 32 bytes to avoid that all subsequent fields are not on word boundary. Then loading a u8 from a 32 byte location is actually more expensive than just the plain full word ;-)

PR #84, opened 2022-02-28
[move-to-yul] Refactor dispatcher to support solidity signature

rahxephon89:
## Motivation

To support interoperability between Move contracts and contracts written in other languages such as Solidity, external signatures need be defined. We cannot direct use the function signatures in Move because types are not compatible. Moreover, parameters in a solidity interface with array, struct and mapping types must be specified with a data location. Moreover, function selectors and ABI-related code in the Yul program should be generated based on this signature.

In this PR, we allow users to specify a solidity signature for callable Move functions. The signature is specified as an ascii string in the callable attribute. For instance, the signature of the function [safeTransferFrom](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/abdb20a6bdb1700d58ea9e01b7471dafdef52a68/contracts/token/ERC721/ERC721.sol#L175) in ERC721 can be specified as #[callable(sig=b”safeTransferFrom(address,address,uint256,bytes memory)”)] in the Move program. The function [getApproved](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/abdb20a6bdb1700d58ea9e01b7471dafdef52a68/contracts/token/ERC721/ERC721.sol#L127) can be specified as  #[callable(sig=b”getApproved(uint) returns (address)”)]. 

The compiler would also check whether the user defined signature is compatible with the Move signature. For instance, the range of a solidity uint type must not exceed its Move correspondence. 

If a solidity signature is not explicitly specified, the compiler will translate the signature of the Move compiler into a compatible solidity signature. 

Yes

## Test Plan

Baseline tests are added to the test-dispatcher/signature-parsing-test for testing parsing the signature and checking the type compatibility. Semantic tests for dispatcher will be added in a separated PR.


language/evm/move-to-yul/src/solidity_ty.rs
```diff
@@ -0,0 +1,543 @@
...
+pub struct SoliditySignature {
+    pub sig_name: String,
+    pub para_types: Vec<(SolidityType, SignatureDataLocation)>,
+    pub ret_types: Vec<(SolidityType, SignatureDataLocation)>,
+}
+
+impl fmt::Display for SoliditySignature {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "sig:{} ", self.sig_name)?;
+        for (ty, data_location) in &self.para_types {
```
wrwg, line 30: Would it be difficult to produce here the correct Solidity syntax?

Then it could be easier to translate a Move signature just into this type, and then use display on this type to produce the signature.

wrwg (commented):
Nice!

language/evm/move-to-yul/src/generator.rs
```diff
@@ -334,63 +335,15 @@ impl Generator {
         });
     }
 
-    /// Generate type string for encoding the function signature
-    fn get_evm_type_string(&self, ctx: &Context, ty: &Type) -> String {
```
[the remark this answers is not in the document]
rahxephon89: Sure, I moved some of the code into the solidity_ty module.

language/evm/move-to-yul/src/solidity_ty.rs
rahxephon89: I added a function to generate the solidity signature for selector in the code.

language/evm/move-to-yul/src/context.rs
```diff
@@ -276,78 +276,51 @@ impl<'a> Context<'a> {
...
         let struct_env = self.env.get_struct(struct_id);
         attributes::is_evm_arith_module(&struct_env.module_env) && struct_env.is_native()
     }
 
     /// Check whether ty is a static type in the sense of serialization
-    pub fn abi_is_static_type(&self, ty: &Type) -> bool {
-        use move_model::ty::{PrimitiveType::*, Type::*};
-
-        let conjunction = |tys: &[Type]| {
+    pub fn abi_is_static_solidity_type(&self, ty: &SolidityType) -> bool {
```
wrwg, line 286: I suggest to move this and similar functions (basically everything which mostly works on `SolidityType`) into module solidity_type:

```
impl SolidityType {
  pub fn is_static(&self, ctx: &Context)
  pub fn size(&self) 
  ...
```

I think this will make the code better readable.

Technically, we have *not* done the same for `model::type::Type` because we cannot add impls to a type from another crate. But inside a crate we should maximize this for better readability.

language/evm/move-to-yul/src/generator.rs
```diff
@@ -727,6 +706,23 @@ impl Generator {
                     // Only dispatch callables
                     continue;
                 }
+                let extracted_sig_opt = attributes::extract_callable_signature(fun);
+                let mut sig = ctx.create_default_solidity_signature(fun);
+                if let Some(extracted_sig) = extracted_sig_opt {
+                    let parsed_sig_opt = ctx.parse_into_solidity_signature(&extracted_sig);
```
wrwg, line 712: Parsing a solidity type or signature from a string should also be a member function of `SolidityType` (perhaps get rid of the `impl Context` inside of `solidity_type` and instead pass `ctx: &Context` where you need it?

language/evm/move-to-yul/src/solidity_ty.rs
```diff
@@ -0,0 +1,627 @@
...
+            Ufixed(m, n) => write!(f, "ufixed{}x{}", m, n),
+            Address(_) => f.write_str("address"),
+        }
+    }
+}
+
+// ================================================================================================
+// Pretty print for SolidityType
+
+impl fmt::Display for SolidityType {
```
wrwg, line 145: nit: still think this display should produce a string which is correct solidity syntax and can be passed into your parser. Nice for testing but also perhaps some point in error messages or similar. If that is a debug print, you would impl the Debug print trait.

language/evm/move-to-yul/src/solidity_ty.rs
```diff
@@ -0,0 +1,627 @@
...
+
+use move_model::{
+    model::{FunctionEnv, QualifiedId, StructId},
+    ty::{PrimitiveType, Type},
+};
+
+use crate::context::Context;
+
+// Solidity signature matching
+static SIG_REG: Lazy<Regex> = Lazy::new(|| {
```
wrwg, line 21: Just a note that you can move the static declaration actually inside the function where you need it, if you want so. This makes it definitely simpler during development as those things are closer together, but also easier to read.

language/evm/move-to-yul/src/solidity_ty.rs
```diff
@@ -0,0 +1,627 @@
...
+                            ret_ty = &stripped_returns_trim[1..stripped_returns_trim.len() - 1];
+                        } else {
+                            parse_error = true;
+                        }
+                    } else {
+                        parse_error = true;
+                    }
+                    if parse_error {
+                        return Err(anyhow!(
+                            "error happens when parsing the return types in the signature"
```
wrwg, line 266: nit: as you decided to put some error messages into const string, perhaps all of them (or none)?

wrwg (changes requested):
Very nice! A few remaining nits, specifically asking for moving more functionality into methods of `SolidityType`, for better code readability.

wrwg (approved):

PR #93, opened 2022-03-03
[evm] fix logical not and other small issues

emmazzz:
## Motivation

This PR fixes an issue regarding logical not. Right now we use Yul's builtin `not` for logical not but that `not` is bitwise so that means both `not(true)` and `not(false)` evaluate to something nonzero. I have changed the implementation to use `iszero` instead and added a test case for it. There are also small bug fixes regarding U256 and a small change in test reporting to include the function's name as well.


wrwg (dismissed):
Nice!

PR #96, opened 2022-03-05
[move-lang] Typed macros: parsing and type checking

wrwg:
(This is a port of [pending Diem PR #1005](https://github.com/diem/diem/pull/10005) from the Diem repo to the Move repo. Mentioned PR was submitted before Move was split off Diem. For the historical discussion see the original version. The port has been created by (a) calling `git format-patch` in the Diem repo (b) calling `git am -3 < patch-file` in the Move repo, so it should be save following the same schema as general git rebasing does.)

This PR realizes the 1st step towards an alternative, high fidelity implementation of user level macros, in contrast to compiler macros suggested in [Diem PR #9930](https://github.com/diem/diem/pull/10005).

In this model, macros can be declared using a new style of function declaration:

```
    public macro foreach<T>(v: &vector<T>, action: |&T|) {
        let i = 0;
        while (i < Vector::length(v)) {
            action(Vector::borrow(v, i));
            i = i + 1;
        }
    }
```

This PR implements parsing and type checking of this style of macros up to the typing AST. Macro expansion is not yet implemented and intended to be added in a followup.

The biggest change is enabling function types, lambda expressions, and calls of function values for naming and typing. Those had been supported before for the spec language, but weren't supported in move-lang beyond expansion. In the design, we represent function types by a new `BuiltinTypeName_::Fun` which allows to minimize changes to the code, as no new type variant needs to be introduced, and existing functionality for inference et. al can be adopted with minimal changes. This consistent e.g. with the treatment of function types in Rust, where they are just type constructors with some syntactic sugar.

A new test case has been added at [lambda.move](https://github.com/diem/move/blob/70cddf1abc124fcc9f606350f520feda829343e7/language/move-compiler/tests/move_check/typing/lambda.move) which shows usage and intends to cover all type check failure scenarios.

After this PR, to complete this style of macros, we still need to:

- Implement actual expansion of macros. This is intended to happen in a new phase after typing.
- Implement representation of macros in compiled, persisted modules. We need to be able to fetch the macro definition from an imported, precompiled module.
- Implement techniques (in the prover) to refer to specifications of passed lambda arguments to macros. (See https://github.com/diem/move/blob/main/language/move-prover/tests/sources/functional/macro_verification.move for what we want to do here.)

## Motivation

Macros

Yes

## Test Plan

Added new tests

### DIFF
--- a/language/move-compiler/src/hlir/ast.rs
+++ b/language/move-compiler/src/hlir/ast.rs
@@ -437,6 +437,7 @@ impl BaseType_ {
                 )
                 .unwrap()
             }
+            Fun => panic!("ICE unexpected function type"),
         };
         let n = sp(loc, TypeName_::Builtin(sp(loc, b_)));
         sp(loc, BaseType_::Apply(kind, n, ty_args))
--- a/language/move-compiler/src/naming/ast.rs
+++ b/language/move-compiler/src/naming/ast.rs
@@ -158,7 +158,7 @@ pub type TypeName = Spanned<TypeName_>;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
 pub enum Variance {
-    CoVariant,
+    Covariant,
     ContraVariant,
 }
 
@@ -394,7 +394,7 @@ impl BuiltinTypeName_ {
             // Function variance: given g: T1 -> R1 and f: T2 -> R2, then
             // f can substitute g if T2 >= T1 && R1 <= R2
             BuiltinTypeName_::Fun if pos < arity - 1 => Variance::ContraVariant,
-            _ => Variance::CoVariant,
+            _ => Variance::Covariant,
         }
     }
 }
@@ -403,7 +403,7 @@ impl TypeName_ {
     pub fn variance(&self, pos: usize, arity: usize) -> Variance {
         match self {
             TypeName_::Builtin(bn) => bn.value.variance(pos, arity),
-            _ => Variance::CoVariant,
+            _ => Variance::Covariant,
         }
     }
 }
--- a/language/move-compiler/src/naming/ast.rs
+++ b/language/move-compiler/src/naming/ast.rs
@@ -102,6 +102,7 @@ pub type FunctionBody = Spanned<FunctionBody_>;
 #[derive(PartialEq, Debug, Clone)]
 pub struct Function {
     pub attributes: Attributes,
+    pub is_macro: bool,
     pub visibility: Visibility,
     pub signature: FunctionSignature,
     pub acquires: BTreeMap<StructName, Loc>,
@@ -140,6 +141,8 @@ pub enum BuiltinTypeName_ {
     Vector,
     // bool
     Bool,
+    // Function (last type arg is result type)
+    Fun,
 }
 pub type BuiltinTypeName = Spanned<BuiltinTypeName_>;
 
@@ -152,6 +155,12 @@ pub enum TypeName_ {
     ModuleType(ModuleIdent, StructName),
 }
 pub type TypeName = Spanned<TypeName_>;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+pub enum Variance {
+    CoVariant,
+    ContraVariant,
+}
 
 #[derive(Debug, Hash, Eq, PartialEq, Ord, PartialOrd, Copy, Clone)]
 pub struct TParamID(pub u64);
@@ -225,9 +234,11 @@ pub enum Exp_ {
     ModuleCall(
         ModuleIdent,
         FunctionName,
+        bool,
         Option<Vec<Type>>,
         Spanned<Vec<Exp>>,
     ),
+    VarCall(Var, Spanned<Vec<Exp>>),
     Builtin(BuiltinFunction, Spanned<Vec<Exp>>),
     Vector(Loc, Option<Type>, Spanned<Vec<Exp>>),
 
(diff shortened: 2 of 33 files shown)


wrwg:
One thought here, after coming back to this 3 month later. I think we should rename `macro` to `template`.

language/move-compiler/src/naming/ast.rs
```diff
@@ -152,6 +155,12 @@ pub enum TypeName_ {
 }
 pub type TypeName = Spanned<TypeName_>;
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+pub enum Variance {
+    CoVariant,
```
sblackshear, line 160:
```suggestion
    Covariant,
```

language/move-compiler/src/naming/ast.rs
```diff
@@ -152,6 +155,12 @@ pub enum TypeName_ {
...
+    ContraVariant,
```
sblackshear, line 161:
```suggestion
    Contravariant,
```

language/move-compiler/src/naming/ast.rs
```diff
@@ -361,15 +374,35 @@ impl BuiltinTypeName_ {
...
             B::Vector => vec![AbilitySet::empty()],
+            B::Fun => (0..arity).map(|_| AbilitySet::empty()).collect(),
+        }
+    }
+
+    pub fn variance(&self, pos: usize, arity: usize) -> Variance {
+        match self {
+            // Function variance: given g: T1 -> R1 and f: T2 -> R2, then
+            // f can substitute g if T2 >= T1 && R1 <= R2
+            BuiltinTypeName_::Fun if pos < arity - 1 => Variance::ContraVariant,
```
sblackshear, line 395:
```suggestion
            BuiltinTypeName_::Fun if pos + 1< arity => Variance::ContraVariant,
```

to avoid underflow if `arity` is 0?

language/move-compiler/src/typing/core.rs
```diff
@@ -1335,13 +1352,20 @@ fn make_tparams(
...
+        .enumerate()
+        .map(|(i, (vloc, constraint))| {
             let tvar = make_tvar(context, vloc);
             context.add_ability_set_constraint(loc, None::<String>, tvar.clone(), constraint);
             match &case {
-                TVarCase::Single(msg) => context.add_single_type_constraint(loc, msg, tvar.clone()),
+                TVarCase::SingleButLast(_) if i + 1 == arity => {
+                    // Last arg (return type of a function): do not add any constraint
+                }
+                TVarCase::SingleButLast(msg) | TVarCase::Single(msg) => {
```
sblackshear, line 1366: What does `SingleButLast` mean when it's not the return type? Would it be possible to call this enum variant `TVarCase::ReturnType` and raise an error when it's not the return type?

sblackshear (dismissed):
LGTM, thought I am not the foremost expert on this part of the codebase...

Also supportive of the `macro` -> `template` renaming as well if it's not too much trouble, but don't mind the macro name either.

tnowacki:
I don't really like either name. But if I had to pick, I'd pick macro. It is probably just a personal bias, but "template" to me always brings up thoughts of C++ templates, which these really aren't. "macro" to me feels more blobby, undefined, and general.

language/move-compiler/src/naming/ast.rs
wrwg: That is the standard rule for variance of function types. Input arguments are contravariant, output arguments covariant. Variance is about substitutability/assignability: you can replace with a subtype for covariance, and with a supertype for contra-variance.  It is needed for the `join` function.

If you want to treat function types just as regular type constructors, what I'm doing here (and is common in multiple languages e.g. C# and I bet also internally in Rust), you need to introduce the concept of variance to define the subtype relation.

language/move-compiler/src/typing/core.rs
wrwg: I renamed this to `Function` for clarity. The meaning of it is as the code says: for any argument which is not a return value, it behaves like Single, for the return argument it is unconstrained.

language/move-compiler/src/hlir/ast.rs
```diff
@@ -436,6 +436,7 @@ impl BaseType_ {
                 )
                 .unwrap()
             }
+            Fun => panic!("ICE unexpected function type"),
```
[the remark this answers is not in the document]
wrwg: > I don't understand why an explicit variance enum was added. Variance already appears through the fact that `join` cares about the order of arguments when it has `TypingCase::Subtype`
> 
> You should be able to express everything you need for variance by flipping the arguments

You are really comparing in join (A -> B) <: (A' -> B'), so the variance is needed if you do not want to special case the function type constructor in join. This can indeed be done, but I find it cleaner to just have a concept of variance.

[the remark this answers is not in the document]
wrwg: > will this be caught by the unimplemented diag?

Should be, otherwise there is in fact an internal error.

language/move-compiler/src/parser/syntax.rs
```diff
@@ -1759,7 +1777,7 @@ fn parse_struct_type_parameters(
 
 // Parse a function declaration:
 //      FunctionDecl =
-//          "fun"
+//          ( "macro" | "fun" )
```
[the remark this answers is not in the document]
wrwg: Done

wrwg:
@tnowacki 
> What dos the arity have to do with variance at all? Very confused (maybe I will be less confused as I keep reading)

This function returns the variance of the ith parameter of the type constructor (type name). 

In Scala, e.g, the variance can be declared at user level (`T[+A, -B]`). I just thought having this as a general concept here might be handy, and it seems to simplify the implementation of function types.

PR #99, opened 2022-03-06
[tools] Compile in release mode for CI testing (+ enable move-cli release builds)

wrwg:
PR #95 is experiencing stack overflows in some unrelated tests. It is well known that Rust debug builds have a large stack consumption. It is also reported the M1 macs have tighter restrictions on stack allocation.       

Our CI tests were until now run in debug mode. This PR switches this to `--release`. However, it turned out that the move-cli and related packages cannot operate in release mode:

- The `target/debug` path was hardwired into the test code.
- The Move VM does not generate trace/debug info when compiled in release mode (which is needed for the testing framework)

Both those problems have been fixed, and we are able now  to run the Move CLI, package system, and unit tests in release mode. Notice that the CLI's own tests execute approx. 10x faster in release mode, so even if compilation is slower, overall CI time should be improved, and also turn-around times for CLI users.

Yes

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -137,7 +137,7 @@ jobs:
           key: ${{ needs.prepare.outputs.changes-target-branch }}
       - name: run unit tests
         run: |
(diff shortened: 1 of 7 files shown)


sblackshear (approved):
Thanks for debugging/fixing this!

PR #108, opened 2022-03-08
A possible model of Jazz continuations in plain Move

wrwg:
This assumes attributes would trigger generation of derived functions for continuations. It passes the compiler and can be handled by the current language.


wrwg:
@shazqadeer added a version of the counter problem using futures. Futures lead to a slightly more natural programming style, as they avoid the reversion of control flow via the continue_with parameter. Instead, the primitive `followed_by` is used to create sequencing.

PR #111, opened 2022-03-09
Provide an option for using 32 byte address

areshand:
## Motivation

This PR is to provide an option for enabling 32 byte address as there could be more addresses needed for future aptos application scenarios.

The initial account authenticator, roughly the hash of the pubic keys for an account, represent the account address at the time of creation. At which point, a user can change their account authenticator to whatever they like. Right now, this is represented by 16 bytes, as a result, the only way to securely create these accounts without some sort of land grabbing is to require proof of key ownership. This is very unnatural in web3 space, where accounts can often be created as a by-product of another operation. In order to support that behavior, we must expand our account addresses to be back at 32-bytes.

Yes

## Test Plan

cargo test -p move-core-types passed

### DIFF
--- a/language/move-core/types/Cargo.toml
+++ b/language/move-core/types/Cargo.toml
@@ -31,5 +31,6 @@ serde_json = "1.0.64"
 
 [features]
 address20 = []
+address32 = []
 default = []
 fuzzing = ["proptest", "proptest-derive"]
--- a/language/move-core/types/src/account_address.rs
+++ b/language/move-core/types/src/account_address.rs
@@ -18,7 +18,13 @@ impl AccountAddress {
 
     /// The number of bytes in an address.
     /// Default to 16 bytes, can be set to 20 bytes with address20 feature.
-    pub const LENGTH: usize = if cfg!(feature = "address20") { 20 } else { 16 };
+    pub const LENGTH: usize = if cfg!(feature = "address20") {
+        20
+    } else if cfg!(feature = "address32") {
+        32
+    } else {
+        16
+    };
 
     /// Hex address: 0x0
     pub const ZERO: Self = Self([0u8; Self::LENGTH]);


wrwg:
Can't open the issue.

There is a proposal in the design tracker to standardize on 20 bytes. Would that not be enough? 20 bytes it Eth standard, and I think it would be pretty important to align for interoperability.

davidiw:
> Can't open the issue.
> 
> There is a proposal in the design tracker to standardize on 20 bytes. Would that not be enough? 20 bytes it Eth standard, and I think it would be pretty important to align for interoperability.

@wrwg, agreed on interoperability, but would like to know more of what that means, I think this is more Move ecosystem, right?

In the meantime, we saw this pull: https://github.com/diem/move/pull/10 where Pontem is using 32-bytes. Why 32-bytes?
* Vitalik thinks 20-bytes is not enough for security, see [this](https://ethereum-magicians.org/t/increasing-address-size-from-20-to-32-bytes/5485)
* 32-bytes maps to the actual authentication token used in Diem so it doesn't require any bit or byte manipulation
* 32-bytes is the same standard as our key, so it is extremely secure, we'd have to start over if that were broken

Any concerns about landing this and figuring out the north star from there?

I know we also have a sync tomorrow, so we can kick start some of the more interesting topics around this (interoperability) then :).

wrwg:
> > Can't open the issue.
> > There is a proposal in the design tracker to standardize on 20 bytes. Would that not be enough? 20 bytes it Eth standard, and I think it would be pretty important to align for interoperability.
> 
> @wrwg, agreed on interoperability, but would like to know more of what that means, I think this is more Move ecosystem, right?

We at Meta feel that the idea of single blockchain which rules them all is something which has slipped far to the horizon. We will likely have to live with many blockchains co-existing for many more years to come. In that context, with interoperability we mean that those chains can interact via "bridges" with each other. One angle of that interoperability are (maybe?) compatible account addresses. Another is a contract language which is capable to run on multiple chains (that is why we are building Move on EVM, and we may also look later on Move on LLVM/Solana).

The hack of extending Move addressed to 20 bytes (not 32 bytes) via a compile time switch is -- well, a hack which we agreed upon as a temporary solution. Arguably we should get rid of it. 

> 
> In the meantime, we saw this pull: #10 where Pontem is using 32-bytes. Why 32-bytes?

You mean 20, 

> 
> * Vitalik thinks 20-bytes is not enough for security, see [this](https://ethereum-magicians.org/t/increasing-address-size-from-20-to-32-bytes/5485)
> * 32-bytes maps to the actual authentication token used in Diem so it doesn't require any bit or byte manipulation
> * 32-bytes is the same standard as our key, so it is extremely secure, we'd have to start over if that were broken
> 
> On the security front, we cannot create unregistered accounts, e.g., where we don't know the entirety of the `auth_key` having an address only if the address is 16 or 20 bytes.

Interesting argument from security side. I BTW also see that Solana uses 32 bytes,.

Unfortunately, 32 bytes will create some headache for us in our compilation scheme to EVM. (The Move to EVM compiler is at a pretty advanced state right now.) We are exploiting the 20 bytes address size by using the remaining 12 bytes in an EVM word for reference tagging. 

> Any concerns about landing this and figuring out the north star from there?

Not necessarily, but is it blocking you right now? When we introduced the 20 bytes, there was a concrete reason which Mysten had for a project they are doing. Let me say like this: if we add the address32 switch, you actually may use some leverage to press that this becomes the default, One could always say Aptos can just use the switch ;-)

> I know we also have a sync tomorrow, so we can kick start some of the more interesting topics around this (interoperability) then :).

Let us chat about this tomorrow.

wrwg:
> The initial account authenticator, roughly the hash of the pubic keys for an account, represent the account address at the time of creation. At which point, a user can change their account authenticator to whatever they like. Right now, this is represented by 16 bytes, as a result, the only way to securely create these accounts without some sort of land grabbing is to require proof of key ownership. 

Just to check whether I understand this right: we want to use 256 bit addresses so we can support a pure authenticator based address creation scheme. Is it because sha128 is not good enough to safely hash the keys? But if so, how could proof of key ownership even help if we stay with 128 bits (different keys can end in hash collision)?

Any insights how chains like Eth get away with this? What process do they use for account creation?

davidiw:
> > > Can't open the issue.
> > > There is a proposal in the design tracker to standardize on 20 bytes. Would that not be enough? 20 bytes it Eth standard, and I think it would be pretty important to align for interoperability.
> > 
> > 
> > @wrwg, agreed on interoperability, but would like to know more of what that means, I think this is more Move ecosystem, right?
> 
> We at Meta feel that the idea of single blockchain which rules them all is something which has slipped far to the horizon. We will likely have to live with many blockchains co-existing for many more years to come. In that context, with interoperability we mean that those chains can interact via "bridges" with each other. One angle of that interoperability are (maybe?) compatible account addresses. Another is a contract language which is capable to run on multiple chains (that is why we are building Move on EVM, and we may also look later on Move on LLVM/Solana).
> 

Completely agreed.


> The hack of extending Move addressed to 20 bytes (not 32 bytes) via a compile time switch is -- well, a hack which we agreed upon as a temporary solution. Arguably we should get rid of it.
> 
> > In the meantime, we saw this pull: #10 where Pontem is using 32-bytes. Why 32-bytes?
> 
> You mean 20,
> 

Probably should have linked to the code, since that was an embedded comment in another PR, see https://github.com/pontem-network/diem/blob/v1.5.1-pre/language/move-core/types/src/account_address.rs#L20

> > * Vitalik thinks 20-bytes is not enough for security, see [this](https://ethereum-magicians.org/t/increasing-address-size-from-20-to-32-bytes/5485)
> > * 32-bytes maps to the actual authentication token used in Diem so it doesn't require any bit or byte manipulation
> > * 32-bytes is the same standard as our key, so it is extremely secure, we'd have to start over if that were broken
> > 
> > On the security front, we cannot create unregistered accounts, e.g., where we don't know the entirety of the `auth_key` having an address only if the address is 16 or 20 bytes.
> 
> Interesting argument from security side. I BTW also see that Solana uses 32 bytes,.
> 
> Unfortunately, 32 bytes will create some headache for us in our compilation scheme to EVM. (The Move to EVM compiler is at a pretty advanced state right now.) We are exploiting the 20 bytes address size by using the remaining 12 bytes in an EVM word for reference tagging.
> 

I imagine these are somewhat orthogonal. Well I reckon there is some benefit for a unified addressing scheme, the ecosystem has quickly out grown it. One cannot take an Ethereum transaction and apply it a non-ethereum compatible blockchain. Likewise a parameter on a smart contract doesn't really care if the blob of bytes implies one thing or another. While there is likely if not definitely some convenience to having a unified addressing scheme, it does come at the cost that if there's a disconnect between key material and address (especially integrity based addresses like bech32) on that chain, users may mistakenly send funds.

There is definitely more to chew on here, but I figured I'd kick the can a little further tonight :).

> > Any concerns about landing this and figuring out the north star from there?
> 
> Not necessarily, but is it blocking you right now? When we introduced the 20 bytes, there was a concrete reason which Mysten had for a project they are doing. Let me say like this: if we add the address32 switch, you actually may use some leverage to press that this becomes the default, One could always say Aptos can just use the switch ;-)
> 

Borderline ... blocking. We definitely want to improve the ergonomics of account management.

> > I know we also have a sync tomorrow, so we can kick start some of the more interesting topics around this (interoperability) then :).
> 
> Let us chat about this tomorrow.

Yes, yes, I suspect there's a lot for us to learn! Have a good night.

davidiw:
> Just to check whether I understand this right: we want to use 256 bit addresses so we can support a pure authenticator based address creation scheme. Is it because sha128 is not good enough to safely hash the keys? But if so, how could proof of key ownership even help if we stay with 128 bits (different keys can end in hash collision)?
> 
> Any insights how chains like Eth get away with this? What process do they use for account creation?

A collision where someone has an authenticator isn't so bad, because the first authenticator to claim it, wins.

A collision where an authenticator is yet to be assigned would be a nightmare. Imagine we had the ability to create accounts as a byproduct of a resource being initialized on it. Ideally the account auth would be the default one for the account. At 16 bytes, we have potential for hash conflicts with enough CPU power. We can imagine that a user might be negligent in claiming their account and so only 16-bytes of authenticator protect their account from another individual claiming it. After a few days, weeks, months, and adversary could likely have computed a key that works.

In Eth, they have 20 bytes of protection. I suspect they're overall confident that this is secure enough for now. Perhaps when we see a decent leap in computation power, they will hurry up. Perhaps around the time my Eth2.0 is available for use again.

I guess another to think about is that purely computing a hash of a public key is insufficient, one must product a private key, derive a public key, and then compute the hash. I'd be curious how much more expensive that is than just trying to find a duplicate hash.

davidiw (dismissed):

PR #116, opened 2022-03-10
[evm] initial implementation of evm-based unit tests

vgao1996:
This is the initial implementation of evm-based Move unit tests. 

What's currently supported:
- Tests without arguments
- `#[expected_failure]`

Future work:
- Removing `#[evm_test]`
- Implementing the missing features:
  - Test arguments
  - Abort/error locations
- Refactoring the unit test framework
  - Generic backend implementation
  - Comparison of results from different backends


language/evm/move-to-yul/src/attributes.rs
```diff
@@ -14,7 +14,8 @@ const EVM_ARITH_ATTR: &str = "evm_arith";
 const PAYABLE_ATTR: &str = "payable";
 const RECEIVE_ATTR: &str = "receive";
 const RECEIVE_FALLBACK_ATTR: &str = "fallback";
-const TEST_ATTR: &str = "evm_test";
+const EVM_TEST_ATTR: &str = "evm_test";
+const TEST_ATTR: &str = "test";
```
tzakian, line 18: What are the semantics of `evm_test` and the difference between this and a normal `test`? Are these tests that should only be run for the EVM?

One worry I have about adding (or continuing support for) the `evm_test` attribute, is that this will need to be added to the compiler so that it can filter out test functions outside of test mode.

language/evm/move-to-yul/src/generator.rs
```diff
@@ -90,7 +91,7 @@ impl Generator {
                 continue;
             }
             for fun in module.get_functions() {
-                if attributes::is_test_fun(&fun) {
+                if attributes::is_evm_test_fun(&fun) {
```
tzakian, line 94: Should this be for both test types?

language/tools/move-unit-test/src/test_runner.rs
```diff
@@ -185,6 +202,47 @@ impl TestRunner {
     }
 }
 
+// TODO: o not expose this to backend implementations
+struct TestOutput<'a, 'b, W> {
```
tzakian, line 206: Nice!

language/tools/move-unit-test/src/test_runner.rs
```diff
@@ -480,4 +508,169 @@ impl SharedTestingConfig {
...
+                || -> TestRunInfo { TestRunInfo::new(function_name.to_string(), duration, 0) };
+
+            // TODO: gas/timeout
+            // TODO: arguments
+            // TODO: locations
+
+            match (test_info.expected_failure.as_ref(), &res.exit_reason) {
+                // Test expected to succeed or abort with a specific abort code, but ran into an internal error.
+                (None | Some(ExpectedFailure::ExpectedWithCode(_)), ExitReason::Revert(_))
+                    if abort_code() == !0u64 =>
```
tzakian, line 586: Is u64 max a differentiated (non-error) return value in the EVM?

Also a nit: May want to use `std::u64::MAX` here instead?

language/evm/move-to-yul/src/attributes.rs
vgao1996: Good question! IIRC the `evm_test` attribute was added to enable us to implement some basic evm-based tests, without having to wait for proper unit test support. I'm planning to kill it very soon -- it could what comes right after this PR.

@wrwg please correct me if I said anything wrong.

language/evm/move-to-yul/src/generator.rs
vgao1996: Not if we are going to remove `#[evm_test]`.

language/tools/move-unit-test/src/test_runner.rs
vgao1996: > Is u64 max a differentiated (non-error) return value in the EVM?
Yes.

> Also a nit: May want to use std::u64::MAX here instead?
Will do.

vgao1996:
I haven't figured out how we can set the feature flag "evm-backend" for tests running on CI, so as a workaround, I made "evm-backend" a default feature so we can have CI coverage while we continue to experiment with the dev tools/setups. Let me know if this sounds like a reasonable approach to you!

language/evm/move-to-yul/src/attributes.rs
wrwg: Yes, it was just a temporary hack.

wrwg (approved):
Wow! This looks very smooth.

PR #119, opened 2022-03-10
[evm] fix small issue with variable naming

emmazzz:
## Motivation

Previously our generated Yul program would error with messages like "`Error: Variable name $linked_dst already taken in this scope`" when we move a multi-layered (>= 3 layers) struct from or to storage. This PR fixes this issue by adding the hash of the type to `$linked_dst` and `$linked_src`. This way there will be no naming collision since we can’t have recursive types. A test is also added to test this.


wrwg (approved):
Nice idea to use the type hash for this.

PR #121, opened 2022-03-11
[evm] support arguments in evm-based unit tests

vgao1996:
This allows arguments (`#[test(a = ...)]`) to be used in EVM-based unit tests. 

It should be noted that right now, the unit test framework itself only supports address arguments (passed to parameters of `signer` or `address` types), so this is all we get for EVM-based unit tests.

### DIFF
--- a/language/evm/move-to-yul/src/generator.rs
+++ b/language/evm/move-to-yul/src/generator.rs
@@ -1,12 +1,11 @@
 // Copyright (c) The Diem Core Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use codespan_reporting::diagnostic::Severity;
-use move_core_types::{identifier::IdentStr, language_storage::ModuleId};
-use std::collections::{BTreeMap, BTreeSet};
-
+use codespan_reporting::{diagnostic::Severity, term::termcolor::Buffer};
 use itertools::Itertools;
+use move_core_types::{identifier::IdentStr, language_storage::ModuleId, value::MoveValue};
 use sha3::{Digest, Keccak256};
+use std::collections::{BTreeMap, BTreeSet};
 
 use move_model::{
     ast::TempIndex,
@@ -93,7 +92,7 @@ impl Generator {
             for fun in module.get_functions() {
                 if attributes::is_evm_test_fun(&fun) {
                     let mut gen = Generator::default();
-                    gen.test_object(&ctx, &fun);
+                    gen.test_object(&ctx, &fun, &[]);
                     res.insert(fun.get_qualified_id(), ctx.writer.extract_result());
                 }
             }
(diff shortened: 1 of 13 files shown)


language/evm/move-to-yul/src/native_functions.rs
```diff
@@ -39,15 +39,12 @@ impl NativeFunctions {
...
-            ctx.env.error(
-                &gen.contract_loc,
-                &format!(
-                    "native function `{}` not implemented for type `{:?}`",
-                    ctx.env
-                        .get_function(fun_id.to_qualified_id())
-                        .get_full_name_str(),
-                    fun_id.inst,
-                ),
+            unimplemented!(
```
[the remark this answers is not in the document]
wrwg: There is a simple rule which one can apply here: if a user input (or any external input) into a program can lead to an error situation, the program should never crash, but gracefully produce an error. Since a user can just declare any kind of native functions, we should better continue to produce an error.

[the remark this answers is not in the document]
vgao1996: > Since a user can just declare any kind of native functions, we should better continue to produce an error.

I see your point and I agree. Now I think this actually reveals a bigger problem: right now, the native function implementations are part of the move-to-yul compiler itself, so it's not something the caller can control, leading to confusion. I think we should
1. move native function implementations into move-stdlib (preferrable gated behind a feature flag)
2. define a native function interface for the move-to-yul compiler and let the caller pass implementations in, similar to how MoveVM handles its native function implementations

cc @emmazzz

language/evm/move-to-yul/src/generator.rs
```diff
@@ -193,8 +218,29 @@ impl Generator {
                 "mstore(${MEM_SIZE_LOC}, memoryguard(${USED_MEM}))"
             );
             self.need_move_function(&fun_id);
+
+            for (idx, arg) in args.iter().enumerate() {
+                emit!(ctx.writer, "let arg{} := ", idx);
```
vgao1996, line 223: It would be great if you guys could double check if I'm using the right naming scheme here for the bindings of the arguments. I used `a0, a1, ...` initially and somehow the compiler complained about conflicting names.

language/evm/move-to-yul/src/generator.rs
```diff
@@ -172,17 +179,35 @@ impl Generator {
...
+                .error(&test.get_loc(), "test functions cannot have return values");
+            return;
+        }
+        if test.get_parameter_count() != args.len() {
             ctx.env.error(
                 &test.get_loc(),
-                "test functions cannot have parameters or return values \
-                    at this point of time",
+                &format!(
+                    "test functions have {} parameters but got {} arguments",
```
wrwg, line 194: test function has {} parameters but {} were provided

language/evm/move-to-yul/src/generator.rs
wrwg: Internal names should be prefixed with `$`. Source level names will be used for locals where they are available (so if you have `a0` in the source, it would potentially conflict). This kind of to make generated code better readable for debugging.

However,  `$t[0-9]+` must be avoided because the stackless bytecode uses it already for temporaries... This is not completely clean, but not that bad that refactoring was worth the effort until now.

language/evm/move-to-yul/src/native_functions.rs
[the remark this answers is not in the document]
wrwg: I think it might be sufficient to have an error message like: "unexpected undefined native function (bug?)"? At least for now.

language/tools/move-unit-test/src/test_runner.rs
```diff
@@ -556,12 +556,30 @@ impl SharedTestingConfig {
...
                 &test_plan.module_id,
                 IdentStr::new(function_name).unwrap(),
-            );
+                &test_info.arguments,
+            ) {
+                Some(yul_code) => yul_code,
+                None => {
+                    // Failed to generate yul code due to some user errors.
+                    // Mark test as failed.
+                    output.fail(function_name);
```
wrwg, line 570: Should we perhaps add here diagnosis from the model?

vgao1996:
@wrwg I made some updates:
1. Added the `$` prefix to argument bindings
2. Reverted the panic back to an error
3. Included (move-to-yul) compiler diagnostics in the test statistics so they are reported to the user
4. Other minor changes as you requested

Let me know if this looks good to you!

wrwg (approved):
Great to see this.

PR #122, opened 2022-03-12
[evm] implement signer native function

emmazzz:
This PR implements native function `Signer::borrow_address` in Yul.


vgao1996 (dismissed):
LGTM. Thanks for implementing this!

BTW, I said `UnitTest::create_signers_for_testing` was implemented but it turned out I was wrong. We still need to get it implemented.

emmazzz:
> LGTM. Thanks for implementing this!
> 
> BTW, I said `UnitTest::create_signers_for_testing` was implemented but it turned out I was wrong. We still need to get it implemented.

Will implement that in a later PR!

PR #124, opened 2022-03-14
[move] Async Move Examples.

wrwg:
This PR adds some examples for asynchronous programming in Move. We use a simple account, where accounts live on different actors. One can deposit, withdraw, and transfer between those accounts. The transfer is the interesting operation because it requires a roundtrip communication between actors: only if deposit on end is successful, should the money be withdrawn on the other.

There are three versions of this example:

- With continuations.
- With futures.
- With plain message passing and an explicit state machine.

## Motivation

Async Move

Yes


language/jmv/examples/account/sources/Runtime.move
```diff
@@ -0,0 +1,12 @@
+/// Basic JMV Runtime support.
+module Jmv::Runtime {
+    /// Returns address of executing actor.
+    public native fun self(): address;
+
+    /// Returns a system time stamp, in nanos.
+    public native fun now(): u64;
```
runtian-zhou, line 7: I don't think it's a good idea to expose `now`. This function call will be non deterministic and might cause state to diverage when retry. A heartbeat transaction to update the `now` as what we did in `DiemFramework` should be the way to go IMHO

language/jmv/examples/account/sources/AccountContinuation.move
```diff
@@ -0,0 +1,75 @@
...
+    #[_generated_cont]
+    native fun cont_xfer_withdraw(v: u64, cont: Cont<Unit>): Cont<Unit>;
+}
+
+/// Jmv supporting modules.
+module Jmv::Cont {
+    use Jmv::Runtime::{Unit, unit};
+
+    /// A type which represents a continuation.
+    native struct Cont<T> has drop;
```
runtian-zhou, line 66: Do you think T could be resource here?

runtian-zhou (commented):
How do you think this programming paradigm would work with Move on EVM? Specifically on the `Jmv::Runtime` package, how do you think those native functiosn should be compiled? I suppose they will be compiled into some event emit and halt?

runtian-zhou:
Another concern that I'm having is about msg deduplication. Do you think the rpc retry should be handled by some external dedup mechanism?

language/jmv/examples/account/sources/Runtime.move
wrwg: Sure, the way how time is represented should be aligned with what we need. However, I'm wondering whether a heartbeat would really help if it comes to retry. It would really need to be somehow a timestamp unique for the entire batch. Or is that what the heartbeat is?

wrwg:
Now contained in #142

PR #127, opened 2022-03-15
Fix serde serialization leading to compile error

areshand:
## Motivation

Fix the issues described here https://github.com/diem/move/issues/126

Yes

## Test Plan

cargo x lint && cargo xfmt && cargo xclippy --all-targets passed
cargo test -p move-core-types passed


sblackshear (dismissed):
/land

davidiw:
I think we need to revert this PR and return to jolestar's PR. This creates an asymmetry in serialization / deserialization which is a disaster waiting to happen.

@areshand / @sblackshear any objections?
I can't tag jolestar.

I'm happy to approve and considering it hasn't been reported and is so new, I think we should be able to safely do this.

PR #136, opened 2022-03-20
Change type of `args` inputs to `execute_function_for_effects` to make it more flexible

gdanezis:
## Motivation

Currently function to enter VM execution take fully owned vectors of `[u8]` only to promptly deserialize them and drop the vectors. This requires VM users to continuously clone, and live with endless allocations / deallocations. 

The solution is to make the `args` input to `execute_function_for_effects` into an `args: Vec<V>`, with `V` being a generic type with `V: Borrow<[u8]>`. And fix called functions to do the same. This is purely a change of type and an addition of `.borrow()` to get a reference when needed, no other functionality changes. 

As a result, those who enjoy passing `Vec<u8>`s around can continue to do so, while others can pass in types that provide a &[u8] without allocations.

### Have you read the [Contributing Guidelines on pull requests]

Yes.

## Test Plan

All tests pass.

### DIFF
--- a/language/move-vm/runtime/src/runtime.rs
+++ b/language/move-vm/runtime/src/runtime.rs
@@ -29,7 +29,7 @@ use move_vm_types::{
     loaded_data::runtime_types::Type,
     values::{Locals, Value},
 };
-use std::collections::BTreeSet;
+use std::{borrow::Borrow, collections::BTreeSet};
 use tracing::warn;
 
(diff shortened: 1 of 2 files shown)


sblackshear (approved):
LGTM! Though the autoformatter needs to be appeased

wrwg (dismissed):
Nice!

PR #139, opened 2022-03-22
[evm] add support for Tables

emmazzz:
## Motivation

This PR implements support for tables in move-to-yul compiler.
Also fixes two small bugs:
- When hashing `(word, group)` in `StorageKey` function, `group` should be shifted left to be at the beginning of the second word.
- `FreezeRef` should be treated like an assignment.


language/evm/move-to-yul/src/context.rs
```diff
@@ -282,6 +282,15 @@ impl<'a> Context<'a> {
...
     }
 
+    /// Returns whether the struct identified by module_id and struct_id is the native Table struct.
+    pub fn is_table(&self, struct_id: QualifiedId<StructId>) -> bool {
+        let struct_env = self.env.get_struct(struct_id);
+        format!(
+            "{}",
+            struct_env.get_name().display(struct_env.symbol_pool())
+        ) == "Table"
+    }
```
wrwg, line 292: nit: pool.string(struct.env.get_name()).as_ref() == "Table"

However, this should definitely also check for module name and address because currently you match any type with simple name Table.

language/evm/move-to-yul/src/tables.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+fn define_empty_fun(gen: &mut FunctionGenerator, ctx: &Context, fun_id: &QualifiedInstId<FunId>) {
+    let key_type = fun_id.inst.get(0).expect("key type");
+
+    // TODO: right now the key to storage is simply hash (table_handle, key), which works when key
+    // is a primitive type. However, this doesn't work when key is a struct or a vector, represented
+    // by a pointer to its data. We need to implement a hash algorithm for structs and vectors before
+    // allowing using structs and vectors as table keys.
+    assert!(
```
wrwg, line 36: We should make a diag here instead of crash.

wrwg (commented):
Some smaller things.

wrwg (approved):

PR #140, opened 2022-03-22
[move-to-yul] Add support for external calls

rahxephon89:
## Motivation

This PR implements making external calls to other contracts. For instance, to call the function isApprovedForAll from a contract that implements IERC721, a native function decorated with the external attribute can be defined as below:

    #[external(sig=b"isApprovedForAll(address,address)returns(bool)"), view]
    public native fun is_approved_for_all(contract: address, account: address, operator: address): bool;

The first parameter of is_approved_for_all is the address of the target contract. When calling is_approved_for_all, an external call to isApprovedForAll in the target contract will be made. If the external call reverts, the current move contract will also revert. Try-call will be implemented in a separated PR.

Yes

## Test Plan

This PR only adds two test cases for baseline tests. Semantic tests will be added later.


wrwg (dismissed):
Awesome!

wrwg (dismissed):

PR #142, opened 2022-03-22
[move] Async Move

wrwg:
This PR realizes an initial implementation of asynchronous Move.

- It adds an attribute deriver for the 'async' flavor in `language/move-compiler/src/attr_derivation/async_deriver.rs`.
- It adds a wrapper around the Move VM, called `AsyncVM`, which supports functionality to create actors and to handle messages (`language/async/move-async-vm`). There are some basic tests for the new VM in the same crate which also illustrate  usage of this API.
- It adds a number of examples for Async Move at `language/async/examples`. However, the features of continuations and futures used in some of the examples aren't yet supported in the implementation, only core message sending.

## Motivation

Async Move

Yes


wrwg:
FYI @sblackshear @tnowacki Some of the things we are doing here (e.g. passing references into Move entrypoints for state) are rather similar to what Sui must do (thanks for adding `*_with_effects`!)

@davidiw @msmouse This PR contains the addition of an 'environment extension' to the Move VM which would also allows to implement tables. You can see how this is used in `language/async/move-async-vm/src/natives.rs`.

wrwg:
This PR has been rebased on the new merged VM native extension model and cleaned up a bit. It is now ready for review again. Ideally, I'd like to land this before I leave for TACAS 3/30.

The PR does not yet have tables integrated, which we clearly want for Async Move. #150 is out in parallel for tables. Once both land, we can integrate tables as an additional extension to the async extension.,

language/extensions/async/move-async-vm/tests/testsuite.rs
```diff
@@ -0,0 +1,403 @@
...
+        let mut tick = 0;
+        // Publish modules.
+        let mut proxy = HarnessProxy { harness: self };
+        let mut session = self.vm.new_session(test_account(), 0, &mut proxy);
+        let mut done = BTreeSet::new();
+        for id in self.module_cache.keys() {
+            self.publish_module(&mut session, id, &mut gas, &mut done)?;
+        }
+        // Initialize actors
+        let mut mailbox: VecDeque<Message> = Default::default();
```
shazqadeer, line 92: It looks like the "init" command does not go through the mailbox. Perhaps in Async Move, you don't think "init" as a command. In Jazz, at the moment, there is nothing special about the init command. Any command could be put into the starting position of the log and executed as the first command.  Just thought I would mention it.

language/extensions/async/move-async-vm/tests/testsuite.rs
```diff
@@ -0,0 +1,403 @@
...
+            ));
+            let result = {
+                let mut proxy = HarnessProxy { harness: self };
+                let session = self.vm.new_session(addr, 0, &mut proxy);
+                session.new_actor(&actor, addr, &mut gas)
+            };
+            self.handle_result(&mut mailbox, result);
+
+            // Put a start message for this actor into the mailbox.
+            let entry_point_id = Identifier::from_str("start")?;
```
shazqadeer, line 107: "start" is the name of init function?

language/extensions/async/move-async-vm/tests/testsuite.rs
```diff
@@ -0,0 +1,403 @@
...
+            let hash = actor_metadata::message_hash(&actor, &entry_point_id);
+            mailbox.push_back((addr, hash, vec![]));
+        }
+
+        // Handle messages until the mailbox is empty.
+        while let Some((actor, message_hash, args)) = mailbox.pop_front() {
```
shazqadeer, line 113: Nice!

shazqadeer (approved):
Indeed, studying testsuite.rs was helpful. It looks like 
1. user writes Move code in a directory as a bunch of files
2. those files and their dependencies are compiled
3. AsyncVM instance is created to load the compiled modules and actors
4. Round-robin execution of the actors until mailbox (across all actors) becomes empty.

Did I get that right?

wrwg:
> Indeed, studying testsuite.rs was helpful. It looks like
> 
> 1. user writes Move code in a directory as a bunch of files
> 2. those files and their dependencies are compiled
> 3. AsyncVM instance is created to load the compiled modules and actors
> 4. Round-robin execution of the actors until mailbox (across all actors) becomes empty.
> 
> Did I get that right?

Yes, that is the basic picture of what would happen for one actor. Indeed the scheduling might be getting more complex in real world (e.g. batches of parallel messages), as the mailbox will be persisted.

language/extensions/async/examples/account/README.md
```diff
@@ -0,0 +1,17 @@
+This package contains examples for programming patterns in Async Move (code name AMV). We use a simple
+account, where accounts live on different actors. One can deposit, withdraw, and transfer
+between those accounts. The transfer is the interesting operation because it requires a roundtrip
+communication between actors: only if deposit on end is successful, should the money be withdrawn
+on the other.
+
+Note that there is a conceptual bug in this solution: since multiple transfers can be initiated simultaneously,
+but the money is not withdrawn before one finishes, the account balance could get negative (similar as the
+reentrancy problem in Solidity). We chose to ignore this for the sake of illustrating the communication patterns.
```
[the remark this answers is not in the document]
wrwg: Yes, both are quite real problems to deal with

language/extensions/async/move-async-vm/tests/testsuite.rs
wrwg: The difference might be that in Move, you cannot have a struct with default values initialized. So you can't have `#[message] fun init(this: &mut Actor)`. On the other hand, the inits are restricted and can't send any messages. So to simulate the Jazz model, one would have `init` just set defaults in the state, and then some `#[message] fun start.`

language/extensions/async/move-async-vm/tests/testsuite.rs
wrwg: This is just a convention for this test suite. Moving forward, we have to see whether we need a default mechanism for `start`.

shazqadeer (dismissed):

PR #144, opened 2022-03-23
[move-vm] Extension mechanism for VM state

wrwg:
[This PR has been split from #142 for better transparency of this new mechanism. There are a few more functions added to the VM which are useful helpers for #142 but have no obvious need in this PR; I left them here to not complicate the split too much.]

This adds an extension mechanism to the Move VM which allows native functions to access some state defined by the execution environment. The mechanism is based on the Rust `Any` type. New variantes of the execute functions in the session (`execute_XXX_with_env_ext`) which allow to pass in the extension which can then be consumed by natives.

No tests are in this PR as the VM doesn't have its own test suite but is tested indirectly via other crates; #142 contains a number of tests for this.

## Motivation

Extensible VM

Yes

### DIFF
--- a/language/move-binary-format/src/errors.rs
+++ b/language/move-binary-format/src/errors.rs
@@ -171,6 +171,8 @@ impl VMError {
     }
 }
 
+impl std::error::Error for VMError {}
+
 #[derive(Debug, Clone)]
 pub struct PartialVMError {
     major_status: StatusCode,
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -660,6 +660,8 @@ pub enum StatusCode {
     CALL_STACK_OVERFLOW = 4021,
     VM_MAX_TYPE_DEPTH_REACHED = 4024,
     VM_MAX_VALUE_DEPTH_REACHED = 4025,
+    VM_EXTENSION_ERROR = 4026,
+
 
     // A reserved status to represent an unknown vm status.
     // this is std::u64::MAX, but we can't pattern match on that, so put the hardcoded value in
--- a/language/move-vm/runtime/src/native_functions.rs
+++ b/language/move-vm/runtime/src/native_functions.rs
@@ -15,6 +15,7 @@ use move_vm_types::{
     natives::function::NativeResult, values::Value,
 };
 use std::{
+    any::{Any, TypeId},
     collections::{HashMap, VecDeque},
     fmt::Write,
 };
@@ -23,6 +24,24 @@ pub type NativeFunction =
     fn(&mut NativeContext, Vec<Type>, VecDeque<Value>) -> PartialVMResult<NativeResult>;
 
 pub type NativeFunctionTable = Vec<(AccountAddress, Identifier, Identifier, NativeFunction)>;
+
+pub fn make_table(
+    addr: AccountAddress,
+    elems: &[(&str, &str, NativeFunction)],
+) -> NativeFunctionTable {
+    elems
+        .iter()
+        .cloned()
+        .map(|(module_name, func_name, func)| {
+            (
+                addr,
+                Identifier::new(module_name).unwrap(),
+                Identifier::new(func_name).unwrap(),
+                func,
+            )
+        })
+        .collect()
+}
 
 pub(crate) struct NativeFunctions(
     HashMap<AccountAddress, HashMap<String, HashMap<String, NativeFunction>>>,
@@ -62,6 +81,40 @@ pub struct NativeContext<'a> {
     data_store: &'a mut dyn DataStore,
     gas_status: &'a GasStatus<'a>,
     resolver: &'a Resolver<'a>,
+    extensions: &'a mut NativeContextExtensions,
+}
+
+/// A data type to represent a heterogeneous collection of extensions which are available to
+/// native functions. A reference to this is passed into the session function execution
+/// entry points.
+#[derive(Default)]
+pub struct NativeContextExtensions {
+    map: HashMap<TypeId, Box<dyn Any>>,
+}
+
+impl NativeContextExtensions {
+    pub fn add<T: Any>(&mut self, ext: T) {
+        assert!(
+            self.map.insert(TypeId::of::<T>(), Box::new(ext)).is_none(),
+            "multiple extensions of the same type not allowed"
+        )
+    }
+
+    pub fn get<T: Any>(&self) -> &T {
+        self.map
+            .get(&TypeId::of::<T>())
+            .expect("dynamic typing error")
+            .downcast_ref::<T>()
+            .unwrap()
+    }
+
+    pub fn get_mut<T: Any>(&mut self) -> &mut T {
+        self.map
+            .get_mut(&TypeId::of::<T>())
+            .expect("dynamic typing error")
+            .downcast_mut::<T>()
+            .unwrap()
+    }
 }
 
 impl<'a, 'b> NativeContext<'a> {
@@ -70,12 +123,14 @@ impl<'a, 'b> NativeContext<'a> {
         data_store: &'a mut dyn DataStore,
         gas_status: &'a mut GasStatus,
         resolver: &'a Resolver<'a>,
+        extensions: &'a mut NativeContextExtensions,
     ) -> Self {
         Self {
             interpreter,
             data_store,
             gas_status,
             resolver,
+            extensions,
         }
     }
 }
@@ -114,5 +169,13 @@ impl<'a> NativeContext<'a> {
             Err(e) if e.major_status().status_type() == StatusType::InvariantViolation => Err(e),
             Err(_) => Ok(None),
         }
+    }
+
+    pub fn extensions(&self) -> &NativeContextExtensions {
+        self.extensions
+    }
+
+    pub fn extensions_mut(&mut self) -> &mut NativeContextExtensions {
+        self.extensions
     }
 }
--- a/language/move-vm/runtime/src/session.rs
+++ b/language/move-vm/runtime/src/session.rs
@@ -3,7 +3,9 @@
 
 use std::borrow::Borrow;
 
-use crate::{data_cache::TransactionDataCache, runtime::VMRuntime};
+use crate::{
+    data_cache::TransactionDataCache, native_functions::NativeContextExtensions, runtime::VMRuntime,
+};
 use move_binary_format::errors::*;
 use move_core_types::{
     account_address::AccountAddress,
@@ -14,7 +16,7 @@ use move_core_types::{
     resolver::MoveResolver,
     value::MoveTypeLayout,
 };
-use move_vm_types::gas_schedule::GasStatus;
+use move_vm_types::{data_store::DataStore, gas_schedule::GasStatus};
 
 pub struct Session<'r, 'l, S> {
     pub(crate) runtime: &'l VMRuntime,
@@ -69,14 +71,39 @@ impl<'r, 'l, S: MoveResolver> Session<'r, 'l, S> {
         args: Vec<Vec<u8>>,
         gas_status: &mut GasStatus,
     ) -> VMResult<Vec<Vec<u8>>> {
-        self.runtime.execute_function(
+        let no_ext = NativeContextExtensions::default();
+        self.execute_function_with_extensions(
             module,
             function_name,
             ty_args,
             args,
-            &mut self.data_cache,
             gas_status,
+            no_ext,
         )
+        .map(|(_, v)| v)
+    }
+
+    /// Same as `execute_function`, but with native context extensions.
+    pub fn execute_function_with_extensions(
+        &mut self,
+        module: &ModuleId,
+        function_name: &IdentStr,
+        ty_args: Vec<TypeTag>,
+        args: Vec<Vec<u8>>,
+        gas_status: &mut GasStatus,
+        mut extensions: NativeContextExtensions,
+    ) -> VMResult<(NativeContextExtensions, Vec<Vec<u8>>)> {
+        self.runtime
+            .execute_function(
+                module,
+                function_name,
+                ty_args,
+                args,
+                &mut self.data_cache,
+                gas_status,
+                &mut extensions,
+            )
+            .map(|v| (extensions, v))
     }
 
     /// Execute `module`::`fuction_name`<`ty_args`>(`args`) and return the effects in
@@ -93,7 +120,7 @@ impl<'r, 'l, S: MoveResolver> Session<'r, 'l, S> {
     /// used to manufacture `signer`'s or `Coin`'s from raw bytes. It is the respmsibility of the
     /// caller (e.g. adapter) to ensure that this power is useed responsibility/securely for its use-case.
     pub fn execute_function_for_effects<V>(
-        mut self,
+        self,
         module: &ModuleId,
         function_name: &IdentStr,
         ty_args: Vec<TypeTag>,
@@ -103,7 +130,33 @@ impl<'r, 'l, S: MoveResolver> Session<'r, 'l, S> {
     where
         V: Borrow<[u8]>,
     {
+        let no_ext = NativeContextExtensions::default();
+        self.execute_function_for_effects_with_extensions(
+            module,
+            function_name,
+            ty_args,
+            args,
+            gas_status,
+            no_ext,
+        )
+        .1
+    }
+
+    /// Same as `execute_function_for_effects`, but with native context extensions.
+    pub fn execute_function_for_effects_with_extensions<V>(
+        mut self,
+        module: &ModuleId,
+        function_name: &IdentStr,
+        ty_args: Vec<TypeTag>,
+        args: Vec<V>,
+        gas_status: &mut GasStatus,
+        mut extensions: NativeContextExtensions,
+    ) -> (Option<NativeContextExtensions>, ExecutionResult)
+    where
+        V: Borrow<[u8]>,
+    {
         let gas_budget = gas_status.remaining_gas().get();
+
         let execution_res = self.runtime.execute_function_for_effects(
             module,
             function_name,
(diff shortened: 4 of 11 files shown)


language/move-vm/runtime/src/native_functions.rs
```diff
@@ -24,6 +25,26 @@ pub type NativeFunction =
...
+                addr,
+                Identifier::new(module_name).unwrap(),
+                Identifier::new(func_name).unwrap(),
+                func,
+            )
+        })
+        .collect()
+}
+
+pub type NoExecEnv = ();
```
damirka, line 46: Hey @wrwg! Is this type ever used? Or is added to be used in the future?

wrwg: Could catch, removed.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -1774,6 +1775,18 @@ impl Vector {
         self.unpack(type_param, 0)?;
         Ok(())
     }
+
+    pub fn to_blob(self) -> PartialVMResult<Vec<u8>> {
```
runtian-zhou, line 1779: Why is this change needed?

runtian-zhou (commented):
I love this idea in general. This would allow us to have extensible stateful native functions. Code here seems good to me but I'll leave the call to @vgao1996 for approval.

language/move-vm/types/src/values/values_impl.rs
wrwg: Some functions are here because they are needed for #142 but in fact that one is not longer needed, so I removed it.

language/move-core/types/src/language_storage.rs
```diff
@@ -124,6 +124,12 @@ impl Display for ModuleId {
     }
 }
 
+impl ModuleId {
+    pub fn short_str_losless(&self) -> String {
```
zekun000, line 128: typo `losless`

wrwg: Fixed

tnowacki (commented):
Just saw this now, will try to make a review tomorrow

wrwg:
I updated this PR to instead of using a single `any::Any` type for extensions, use a collection of such values which is type indexed. This appears necessary so we can have extensions live side-by-side, e.g. 'table' together with 'async'.

language/move-binary-format/src/errors.rs
```diff
@@ -169,8 +169,30 @@ impl VMError {
             offsets,
         )
     }
+
+    pub fn to_partial(self) -> PartialVMError {
```
tnowacki, line 173: This function was sort of intentionally missing. What case did you need this for?

wrwg: There was a case where I needed to call a public API from internal, but I don't recall which, so I removed this for now. May need to bring it back when I need in the async PR. In general, I think the types here (and elsewhere) are overly strict, but that is just my taste.

language/move-core/types/src/language_storage.rs
```diff
@@ -124,6 +124,12 @@ impl Display for ModuleId {
     }
 }
 
+impl ModuleId {
+    pub fn short_str_lossless(&self) -> String {
+        format!("@{}::{}", self.address.short_str_lossless(), self.name)
```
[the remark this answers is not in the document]
wrwg: For having logs with ModuleIds which aren't unreadable because of mountains of zeros. Again, I removed this and can bring it back in the other PR.

language/move-vm/runtime/src/native_functions.rs
```diff
@@ -24,6 +25,24 @@ pub type NativeFunction =
 
 pub type NativeFunctionTable = Vec<(AccountAddress, Identifier, Identifier, NativeFunction)>;
 
+pub fn make_table(
+    addr: AccountAddress,
+    elems: &[(&str, &str, NativeFunction)],
```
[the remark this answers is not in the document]
wrwg: Well, I tried, but it appears I'm hitting a Snafu with a known issue in Rust about conversion from fn item to fn pointer, e.g. discussed here: https://users.rust-lang.org/t/difference-between-fn-pointer-and-fn-item/32642. I got impatient with figuring the solution so I leave it as is.

language/move-vm/runtime/src/native_functions.rs
```diff
@@ -62,6 +81,40 @@ pub struct NativeContext<'a> {
...
     resolver: &'a Resolver<'a>,
+    extensions: &'a mut NativeContextExtensions,
+}
+
+/// A data type to represent a heterogeneous collection of extensions which are available to
+/// native functions. A reference to this is passed into the session function execution
+/// entry points.
+#[derive(Default)]
+pub struct NativeContextExtensions {
+    map: HashMap<TypeId, Box<dyn Any>>,
```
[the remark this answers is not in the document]
wrwg: Partitioning by module seems to be too restrictive. For example, in the async extension, we have a state shared by a set of natives from different modules. The type-based partitioning seems save because the type is fully owned by one given extension, and can be even private to that extension.

language/move-core/types/src/vm_status.rs
```diff
@@ -660,6 +660,8 @@ pub enum StatusCode {
     CALL_STACK_OVERFLOW = 4021,
     VM_MAX_TYPE_DEPTH_REACHED = 4024,
     VM_MAX_VALUE_DEPTH_REACHED = 4025,
+    VM_EXTENSION_ERROR = 4026,
```
vgao1996, line 663: Could you clarify exactly what this error stands for? Just want to understand whether it's supposed to be a user error or an internal error/invariant violation.

language/move-vm/runtime/src/native_functions.rs
```diff
@@ -62,6 +81,40 @@ pub struct NativeContext<'a> {
...
     gas_status: &'a GasStatus<'a>,
     resolver: &'a Resolver<'a>,
+    extensions: &'a mut NativeContextExtensions,
+}
+
+/// A data type to represent a heterogeneous collection of extensions which are available to
+/// native functions. A reference to this is passed into the session function execution
+/// entry points.
+#[derive(Default)]
+pub struct NativeContextExtensions {
```
vgao1996, line 91: An alternative approach is to take in native functions as closures (`dyn FnMut`) as opposed to plain function pointers. The closures can access whatever outer contexts using `Rc`s. I wonder how that approach compares to this.

vgao1996 (commented):
Overall I think this is a neat idea and I like it. I can also see how we can use this to implement events.

IIRC, @tnowacki and I had a related discussion when we were working on extensible native functions, and one concern that was brought up was that the mutation of such external environments breaks the transactional nature of the Move VM. With that being said I think things are different now with more discussions around giving more control to the adapter layer, and that "being transactional" is a property the adapter (not necessarily the Move VM) needs to hold.

language/move-vm/runtime/src/native_functions.rs
wrwg: I think this wont very well. For example, you can't have to closures which access the same mut state, you'd need to use RefCell for that. The current approach works rather fine, I've implemented async with it and tables are also mostly done. Also let me note that type indexed memory like here is a standard design pattern in C++/Move. Its type safe by construction, because you cant apply a downcast with the wrong type.

language/move-vm/runtime/src/native_functions.rs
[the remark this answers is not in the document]
tnowacki: > can be even private to that extension.
Ahhhhh, makes sense, I like it :)

language/move-vm/runtime/src/native_functions.rs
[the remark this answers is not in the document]
tnowacki: Haha, already this should do then

language/move-vm/runtime/src/session.rs
```diff
@@ -68,6 +70,27 @@ impl<'r, 'l, S: MoveResolver> Session<'r, 'l, S> {
...
+
+    /// Same as `execute_function`, but with native context extensions.
+    pub fn execute_function_with_extensions(
+        &mut self,
+        module: &ModuleId,
+        function_name: &IdentStr,
+        ty_args: Vec<TypeTag>,
+        args: Vec<Vec<u8>>,
+        gas_status: &mut GasStatus,
+        extensions: &mut NativeContextExtensions,
```
tnowacki, line 93: Hmmm, I'm not sure we want to pass it in this way.

In the extensions here, I wanted to make it clear when to discard/keep the changes here. So I would have expected this being passed back alongside the execution result

tnowacki: Maybe we could pass this in as an owned value? And then only pass it back alongside the change set on success?

wrwg: This makes only sense for the `execute_function_with_effects(self, ...)` version, and I did it for this one based on your feedback.

The other functions and this one are already designed for taking mutable references to handle state. Right now its not only `self`, but also `gas_status` which is treated like this. It's fitting the current APIs to use `&mut` extensions as well.

I hope we can move forward with this PR now. This is blocking other work, and I'm on time pressure as I'm going on a trip soon.

tnowacki: > I hope we can move forward with this PR now. This is blocking other work, and I'm on time pressure as I'm going on a trip soon.

As you know today will be a bit of a chaotic day for me, but I'll try my best to get through this 

> The other functions and this one are already designed for taking mutable references to handle state. Right now its not only self, but also gas_status which is treated like this. It's fitting the current APIs to use &mut extensions as well.

Gas status is a bit different because it needs to be modifying the gas as the interpreter is running.
Similarly, the DataStore is also only mut for module publishing... and events... which it really shouldn't be (which I think long ago we were hoping to fix). There is no ability to otherwise write to global storage with that API, hence the &mut is a bit different there

This NativeContextExtensions is jsut way too general to pass in via &mut in my opinion. I understand you can still do non-transactional things if you start using refcells/rcs/etc, but I want to try to make it obvious what the correct behavior should be here. And I think that means passing it in as an owned value everywhere, and only returning it for functions that also produce a changeset

wrwg: I disagree and don't understand why you are holding up this PR and this nitty detail.

All those APIs work on mutable inputs. Why should extensions be different? What is special about them?

This is a minor stylistic issue you are bringing up here. If you aren't happy with the way how the current APIs are designed (i.e. working on `&mut) feel free to come back and refactor it.

language/move-vm/runtime/src/session.rs
```diff
@@ -93,38 +117,68 @@ impl<'r, 'l, S: MoveResolver> Session<'r, 'l, S> {
...
+                        events,
+                        return_values,
+                        mutable_ref_values,
+                        gas_used,
+                    },
+                ),
+                Err(error) => (extensions, ExecutionResult::Fail { error, gas_used }),
             },
-            Err(error) => ExecutionResult::Fail { error, gas_used },
+            Err(error) => (extensions, ExecutionResult::Fail { error, gas_used }),
```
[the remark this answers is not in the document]
wrwg: I disagree. Extensions can record information about failure, and therefore should be passed back.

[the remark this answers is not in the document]
tnowacki: What does that mean? Like no other part of Move has the ability to do that (outside of maybe gas? if you squint a bit), you either get an abort code, or produce a changeset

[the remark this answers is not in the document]
wrwg: You gave the answer in your reply already: gas is an example. An extension may record info similar as GasStatus. It is a generic concept.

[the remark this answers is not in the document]
vgao1996: Not 100% sure if it applies here, but I think here's another potential application to consider: events.

Different platforms may have different semantics for events. For example, EVM drops all events if the transaction gets reverted, while Diem/Move still emits them in case of failure. As Move evolve into a language targeting multiple platforms, I wonder if we would benefit from some of the extra flexibilities here.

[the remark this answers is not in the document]
tnowacki: Backing up a bit here, its not about &mut or not. It's about what effects are easily observable in the case that the Move program aborts. Currently no changes to the outer state are observable because all changes are local, because the session has a local view of changes to storage, and reveals them only through finish(). It has no ability to write to storage because the DataStore API doesn't expose it.

My worry is that the way you have setup the extensions would be like if the session and the interpreter wrote to storage directly, and then it was up to the adapter to back to those changes out in the case of failure. Instead of what we do now, which is the VM never touches your data directly, and instead produces a set of changes locally and returns them with the changeset

This is exactly what @vgao1996 was referring to:

> IIRC, @tnowacki and I had a related discussion when we were working on extensible native functions, and one concern that was brought up was that the mutation of such external environments breaks the transactional nature of the Move VM. With that being said I think things are different now with more discussions around giving more control to the adapter layer, and that "being transactional" is a property the adapter (not necessarily the Move VM) needs to hold.

I agree with @vgao1996  that we are more in the world now where the adapter has some more responsibilities. But I still want to make sure that doing the wrong thing is impossible/difficult. And doing the right thing is obvious/easy. 

I worry with the current PR here, it would be very difficult for an adapter to keep a transactional view of any data passed around these extensions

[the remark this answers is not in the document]
wrwg: Extensions are -- as the name says -- extensions to the current session execution model. I would expect that a well designed extension would follow exactly the same model as the session: it can be created, then a sequence of mutations happens (via native functions), and then there is a finalize call which consumes the extension and computes the extension changeset. There can be, however, also finalize_on_failure, and that is what this API tries supports.

I'm not in favor of restricting APIs based on speculation about the semantics of future extensions. There maybe many reasons why want to have finalize_on_failure for extensions. Beside some kind of custom gas, there can be some debugging info which one wants to pick from an extension. Why should we restrict this? I did not hear any convincing arguments. And I don't really understand why we spent that much time on this little technical detail. As said, we can always come back and change if we feel it's important enough. Let's build the first actual extensions and possibly iterate, instead of being bogged down on this small technical detail.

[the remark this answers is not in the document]
tnowacki: Missed @vgao1996's comment here

> while Diem/Move still emits them in case of failure.

That seems... odd. I did not know it did that, lol. Feels weird to have an event like "PaymentSent" or something just to load it up and see no such transaction. (we can take the discussion offline as it is not pertinent to this PR, but do you recall why that behavior exists, @vgao1996?)

> instead of being bogged down on this small technical detail.

I'm not trying to do that @wrwg, and I apologize if it is coming across that way. To summarize, the concern I have is that the modifications to this native function storage will get **accidentally** misused by an adapter. I want to make sure that any non-transactional behavior, such as Diem events (apparently), is done so intentionally and purposefully. 

I am trying to make sure the design fits into that line of thinking. My hope is that by passing in an owned copy of the extension to all functions, and only getting it back through endpoints that are designed to be run for side-effects. That there would be some more apparently intentionality in the design. Of course you can always sidestep this by wrapping data in `RC`s, or something similar, inside of the native extension, but you have to at least do some work to do that, thus making it an intentional choice. 
And obviously here we aren't doing much by forcing to have an owned value, and I wish there was a better solution (i've thought of doing things like requiring `Send` or `Sync` on the value but I don't think that helps too much really), but I haven't come up with much other than just the owned value thing.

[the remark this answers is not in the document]
wrwg: >  To summarize, the concern I have is that the modifications to this native function storage will get accidentally misused by an adapter. 

I think this is overdesign at the wrong place. Adapters are not everyday code and there aren't dozens of them. People which write adapters can shoot into there foot in many ways and need to know what they are doing. Restricting them unnecessarily (like you are suggesting here) is not helpful for the future of the evolution of Move. There have been multiple examples brought to your attention where having the extension also after failure is useful: the event semantics, debugging, additional gas model for natives, etc.

Besides, this is a new experimental API and we can iterate as we evolve the extension mechanism. It does not make sense to block this PR based on the arguments you brought so far,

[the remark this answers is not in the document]
wrwg: ... for the sake of moving forward and look at more interesting things, I will adopt the PR to what you suggest, while I disagree anyway. Stay tuned for the next commit.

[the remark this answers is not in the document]
wrwg: The change you are asking for has been pushed now.

tnowacki (dismissed):
> I'm not in favor of restricting APIs based on speculation about the semantics of future extensions. There maybe many reasons why want to have finalize_on_failure for extensions. Beside some kind of custom gas, there can be some debugging info which one wants to pick from an extension. Why should we restrict this? I did not hear any convincing arguments. And I don't really understand why we spent that much time on this little technical detail. As said, we can always come back and change if we feel it's important enough. Let's build the first actual extensions and possibly iterate, instead of being bogged down on this small technical detail.

I think I've just done a bad job explaining myself, and I'm sorry for that. 
I had/have this concern about the way the native extension data will be used, and commented on the owned vs &mut hoping that this might help the data be used in a sane way. It very well might do absolutely nothing. My point is that I have this kind of concern, and I was not trying to say that it *must be solved this way and no other!* I think I just needed a discussion on the sort of usage patterns, but instead we got ratholed on this one specific idea I had, which I'm sorry I might have caused.

I'm approving this but I would expect some more changes here in the future. Looking at how the Session pieces together out after the changes, it might just make more sense to take it in as field as a part of the session? And then pass it back in either case on finish? I'm not sure. I know @vgao1996 might have some opinions as the current decisions here might make his famous "session lifetime issue" better/worse. We should discuss more :)

vgao1996:
@tnowacki your comment was spot on that this was related to the "session lifetime issue" I've been ranting about for ages. 

Here's a short version of my opinion: it is exactly due to this "session lifetime issue" (or our plan to address this soon) that the current decisions don't matter that much:
- With the current (problematic) session implementation, adapter implementors already have to be extra careful with lifetimes so not to shoot themselves in the foot, and requiring them to take care of this one more thing won't really make the situation dramatically worse, especially in the short term when we have a handful of clients depending on it.
- The potential fix to the lifetime issue we've been discussing will likely require us to get rid of the current Session altogether, and revisit a lot of the public APIs of the Move VM. The exact design of the extensions will need to be reevaluated within the new context anyway, and there will be chances for us to correct our design mistakes (if there are any).

PR #145, opened 2022-03-24
[move-package] Use source files for dependencies

tnowacki:
## Motivation

- Source files are now used for dependencies
  - This will allow for source language only features to be used across packages
  - Short term this means finally using the vector bytecodes instead of native functions 
  - Long term this will allow for macros and other cool features
- Relaxed the constraint that named addresses are not interchangeable with their value
  - This restriction made a lot more sense when named addresses weren't required to have bindings for most of the compiler passes
  - With the package system defacto controlling these things, the restriction seems a bit silly
  - I would think about this in terms of rust crates: If you make an two aliases for a given crate in a file, either are fine to use and are interchangeable 
  - We should probably add a warning or just ban raw flat addresses, at least from the package system

## Test Plan

- ran package test
- updated source language tests


tzakian (approved):
Package changes look good to me!

PR #146, opened 2022-03-24
allow EVM-based unit tests to be run through move-cli and fix some issues with the stackless-vm-based ones

vgao1996:
This allows EVM-based unit tests to be run through the move-cli and fixes some issues with the stackless-vm-based ones.

I've added some move-cli tests and manually checked that it can correctly launch the MoveStdlib unit tests. (Some MoveStdlib tests are failing due to `move-to-yul` issues, but that's a separate thing to fix...)

There are still some additional issues we need to sort out, but they are mostly about internal implementations/automated testing and shouldn't prevent you from running the tool manually (via `move-cli package test --evm`).

List of known issues:
1. Paths of some (removed) mv interface files are being passed into the move model builder. Currently I'm working around this by filtering out the paths that contain "mv_interfaces". This is a dirty hack only meant for the short term while I try to figure out the root cause.
2. `move-cli` is producing weird error messages in its own unit tests, preventing me from adding negative test cases. This doesn't seem to be an issue when you run the executable manually, which is actually concerning to me since this seem to imply we have different code paths for tests and manual invocations.
3. ~~The EVM or stackless-vm -based move cli unit tests should be gated behind the "evm-backend" feature flag~~ Done

### DIFF
--- a/language/tools/move-cli/src/package/cli.rs
+++ b/language/tools/move-cli/src/package/cli.rs
@@ -27,9 +27,11 @@ use move_command_line_common::files::{FileHash, MOVE_COVERAGE_MAP_EXTENSION};
 use move_compiler::{
     compiled_unit::{CompiledUnit, NamedCompiledModule},
     diagnostics::{self, codes::Severity},
+    shared::{NumberFormat, NumericalAddress},
     unit_test::{plan_builder::construct_test_plan, TestPlan},
     PASS_CFGIR,
 };
+use move_core_types::account_address::AccountAddress;
 use move_coverage::{
     coverage_map::{output_map_to_file, CoverageMap},
     format_csv_summary, format_human_summary,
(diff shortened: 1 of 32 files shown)


language/tools/move-cli/src/package/cli.rs
```diff
@@ -163,6 +166,12 @@ pub enum PackageCommand {
         /// Collect coverage information for later use with the various `package coverage` subcommands
         #[clap(long = "coverage")]
         compute_coverage: bool,
+
+        /// Use the EVM-based execution backend.
+        /// Does not work with --stackless.
+        #[cfg(feature = "evm-backend")]
+        #[structopt(long = "evm")]
```
wrwg, line 173: How can structopt still be here? We've been switched by a friendly soul to solely use clap 3, which AFAICT includes structopt.

wrwg (approved):
nit about the structopt. If it works you can also fix this in the next PR.

language/tools/move-cli/src/package/cli.rs
vgao1996: I double checked `move-cli`'s Cargo.toml and can confirm that we have indeed switched to clap 3. Maybe they kept `structopt` for backward compatibility reasons? Anyway, I'll get this fixed in a future PR.

PR #148, opened 2022-03-25
[evm] Added unit tests for ERC20 and ERC1155 contracts

junkil-park:
This PR adds some unit tests for ERC20 and ERC1155 contracts, so it
tests out both Emma's Table implementation and Victor's unit testing
support in CLI.

It turns out that the unit testing in move-cli has some issue. See a minimal
example to reproduce the issue in `evm/examples/tests/ERC20Tests.move`.
Instead, this PR adds the unit tests to the `cargo test` testsuite of
the the `move-to-yul` crate.

In addition, this PR:
- implements the "ownable" pattern for ERC1155, and made ERC1155
  mintable
- allows the initial supply for ERC20
- implements the translation of the native function `Evm::self()`

TODO:
- event-related code is commented out because it's not supported in unit testing at the moment.
- `Std::ASCII::is_valid_char` used to use the parameter name `byte` which is incompatible with the `move-to-yul` translation.

## Motivation

To test ERC contracts

yes

## Test Plan

cargo test

### DIFF
--- a/language/evm/examples/sources/ERC1155.move
+++ b/language/evm/examples/sources/ERC1155.move
@@ -49,6 +49,7 @@ module Evm::ERC1155 {
         balances: Table<U256, Table<address, U256>>,
         operatorApprovals: Table<address, Table<address, bool>>,
         uri: String,
+        owner: address, // Implements the "ownable" pattern.
     }
 
     #[create]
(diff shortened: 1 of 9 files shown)


vgao1996 (dismissed):
LGTM

wrwg (dismissed):
Wow! Awesome that we have the first regular unit tests.

vgao1996 (dismissed):

PR #149, opened 2022-03-25
[move-to-yul] Add support for emitting events

rahxephon89:
## Motivation

This PR adds support for emitting events from the move contract. Users can specify an attribute to a struct as events:

    #[event(sig=b"Transfer(address indexed, address,uint256)")]
    struct Transfer {
        from: address,
        to: address,
        value: U256,
    }

Then, the event can be emitted by the emit function:

    #[callable]
    fun do_transfer(from: address, to: address, amount: U256){
        emit(Transfer{from, to, value: amount});
    }

Yes

## Test Plan

Baseline tests are added to the test-events folder. Semantic tests need to be implemented in the transactional tests, which will be added later.


wrwg (dismissed):
Looks great!

PR #150, opened 2022-03-26
[move-lang] Table extension.

wrwg:
This adds tables as an extension to Move, using the new VM extension mechanism.

Tables (and other future extensions) are suggested to live in the `language/extensions` tree. They come with a Move library and a Rust implementation of the table extension data, including changeset representation, and a native function table.

The CLI and unit testing framework has been extended to support extensions. This is controlled by a compile-time feature flag `table-extension`. The mechanism is generic enough to also support future extensions.

A fairly complete test suite has been copied over from the EVM project.  However, while functionality should be covered, Move unit tests do not support testing of change set generation and remote table resolver access, so this aspect stays untested. For this integration tests are needed.

## Motivation

Tables

Yes

### DIFF
--- a/language/extensions/move-table-extension/Move.toml
+++ b/language/extensions/move-table-extension/Move.toml
@@ -1,0 +1,14 @@
+[package]
+name = "MoveTableExtension"
+version = "1.0.0"
+
+[addresses]
+Std = "_"
+Extensions = "_"
+
+[dev-addresses]
+Std = "0x1"
+Extensions = "0x2"
+
+[dependencies]
+MoveStdlib = { local = "../../move-stdlib" }
--- a/language/extensions/move-table-extension/sources/Table.move
+++ b/language/extensions/move-table-extension/sources/Table.move
@@ -1,0 +1,94 @@
+/// Type of large-scale storage tables.
+module Extensions::Table {
+    /// Type of tables
+    struct Table<phantom K, phantom V> has store {
+        handle: u128
+    }
+
+    /// Create a new Table.
+    public fun new<K, V: store>(): Table<K, V> {
+        Table{handle: new_table_handle()}
+    }
+
+    /// Destroy a table. The table must be empty to succeed.
+    public fun destroy_empty<K, V>(table: Table<K, V>) {
+        destroy_empty_box<K, V, Box<V>>(&table);
+        drop_unchecked_box<K, V, Box<V>>(table)
+    }
+
+    /// Add a new entry to the table. Aborts if an entry for this
+    /// key already exists. The entry itself is not stored in the
+    /// table, and cannot be discovered from it.
+    public fun add<K, V>(table: &mut Table<K, V>, key: &K, val: V) {
+        add_box<K, V, Box<V>>(table, key, Box{val})
+    }
+
+    /// Acquire an immutable reference to the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    public fun borrow<K, V>(table: &Table<K, V>, key: &K): &V {
+        &borrow_box<K, V, Box<V>>(table, key).val
+    }
+
+    /// Acquire a mutable reference to the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    public fun borrow_mut<K, V>(table: &mut Table<K, V>, key: &K): &mut V {
+        &mut borrow_box_mut<K, V, Box<V>>(table, key).val
+    }
+
+    /// Returns the length of the table, i.e. the number of entries.
+    public fun length<K, V>(table: &Table<K, V>): u64 {
+        length_box<K, V, Box<V>>(table)
+    }
+
+    /// Returns true if this table is empty.
+    public fun empty<K, V>(table: &Table<K, V>): bool {
+        length(table) == 0
+    }
+
+    /// Acquire a mutable reference to the value which `key` maps to.
+    /// Insert the pair (`key`, `default`) first if there is no entry for `key`.
+    public fun borrow_mut_with_default<K, V: drop>(table: &mut Table<K, V>, key: &K, default: V): &mut V {
+        if (!contains(table, key)) {
+            add(table, key, default)
+        };
+        borrow_mut(table, key)
+    }
+
+    /// Remove from `table` and return the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    public fun remove<K, V>(table: &mut Table<K, V>, key: &K): V {
+        let Box{val} = remove_box<K, V, Box<V>>(table, key);
+        val
+    }
+
+    /// Returns true iff `table` contains an entry for `key`.
+    public fun contains<K, V>(table: &Table<K, V>, key: &K): bool {
+        contains_box<K, V, Box<V>>(table, key)
+    }
+
+    #[test_only]
+    /// Testing only: allows to drop a table even if it is not empty.
+    public fun drop_unchecked<K, V>(table: Table<K, V>) {
+        drop_unchecked_box<K, V, Box<V>>(table)
+    }
+
+    // ======================================================================================================
+    // Internal API
+
+    /// Wrapper for values. Required for making values appear as resources in the implementation.
+    struct Box<V> has key, drop, store {
+        val: V
+    }
+
+    // Primitives which take as an additional type parameter `Box<V>`, so the implementation
+    // can use this to determine serialization layout.
+    native fun new_table_handle(): u128;
+    native fun add_box<K, V, B>(table: &mut Table<K, V>, key: &K, val: Box<V>);
+    native fun borrow_box<K, V, B>(table: &Table<K, V>, key: &K): &Box<V>;
+    native fun borrow_box_mut<K, V, B>(table: &mut Table<K, V>, key: &K): &mut Box<V>;
+    native fun length_box<K, V, B>(table: &Table<K, V>): u64;
+    native fun contains_box<K, V, B>(table: &Table<K, V>, key: &K): bool;
+    native fun remove_box<K, V, B>(table: &mut Table<K, V>, key: &K): Box<V>;
+    native fun destroy_empty_box<K, V, B>(table: &Table<K, V>);
+    native fun drop_unchecked_box<K, V, B>(table: Table<K, V>);
+}
--- a/language/extensions/move-table-extension/src/lib.rs
+++ b/language/extensions/move-table-extension/src/lib.rs
@@ -1,0 +1,606 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! A crate which extends Move by tables.
+//!
+//! See [`Table.move`](../sources/Table.move) for language use.
+//! See [`README.md`](../README.md) for integration into an adapter.
+
+use move_binary_format::errors::{PartialVMError, PartialVMResult};
+use move_core_types::{
+    account_address::AccountAddress,
+    gas_schedule::{GasAlgebra, GasCarrier, InternalGasUnits},
+    value::MoveTypeLayout,
+    vm_status::StatusCode,
+};
+use move_vm_runtime::{
+    native_functions,
+    native_functions::{NativeContext, NativeFunctionTable},
+};
+use move_vm_types::{
+    loaded_data::runtime_types::Type,
+    natives::function::NativeResult,
+    pop_arg,
+    values::{GlobalValue, GlobalValueEffect, Reference, StructRef, Value},
+};
+use once_cell::sync::Lazy;
+use sha3::{Digest, Sha3_256};
+use smallvec::smallvec;
+use std::{
+    cell::RefCell,
+    collections::{btree_map::Entry, BTreeMap, BTreeSet, VecDeque},
+    convert::TryInto,
+    fmt::Display,
+};
+
+// ===========================================================================================
+// Public Data Structures and Constants
+
+/// The representation of a table handle. This is created from truncating a sha3-256 based
+/// hash over a transaction hash provided by the environment and a table creation counter
+/// local to the transaction.
+#[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq)]
+pub struct TableHandle(pub u128);
+
+impl Display for TableHandle {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "T-{:X}", self.0)
+    }
+}
+
+/// A table change set.
+#[derive(Default)]
+pub struct TableChangeSet {
+    pub new_tables: BTreeSet<TableHandle>,
+    pub removed_tables: BTreeSet<TableHandle>,
+    pub changes: BTreeMap<TableHandle, TableChange>,
+}
+
+/// A change of a single table.
+pub struct TableChange {
+    pub entries: BTreeMap<Vec<u8>, Option<Vec<u8>>>,
+}
+
+/// A table resolver which needs to be provided by the environment. This allows to lookup
+/// data in remote storage, as well as retrieve cost of table operations.
+pub trait TableResolver {
+    fn resolve_table_entry(
+        &self,
+        handle: &TableHandle,
+        key: &[u8],
+    ) -> Result<Option<Vec<u8>>, anyhow::Error>;
+
+    fn table_size(&self, handle: &TableHandle) -> Result<usize, anyhow::Error>;
+
+    fn operation_cost(
+        &self,
+        op: TableOperation,
+        key_size: usize,
+        val_size: usize,
+    ) -> InternalGasUnits<GasCarrier>;
+}
+
+/// A table operation, for supporting cost calculation.
+pub enum TableOperation {
+    NewHandle,
+    Destroy,
+    Insert,
+    Borrow,
+    Length,
+    Remove,
+    Contains,
+}
+
+/// The native table context extension. This needs to be attached to the NativeContextExtensions
+/// value which is passed into session functions, so its accessible from natives of this
+/// extension.
+pub struct NativeTableContext<'a> {
+    resolver: &'a dyn TableResolver,
+    txn_hash: u128,
+    table_data: RefCell<TableData>,
+}
+
+pub static ALREADY_EXISTS: Lazy<u64> = Lazy::new(|| unique_sub_status_code(0));
+pub static NOT_FOUND: Lazy<u64> = Lazy::new(|| unique_sub_status_code(1));
+pub static NOT_EMPTY: Lazy<u64> = Lazy::new(|| unique_sub_status_code(2));
+
+// ===========================================================================================
+// Private Data Structures and Constants
+
+/// A structure representing mutable data of the NativeTableContext. This is in a RefCell
+/// of the overall context so we can mutate while still accessing the overall context.
+#[derive(Default)]
+struct TableData {
+    new_tables: BTreeSet<TableHandle>,
+    removed_tables: BTreeSet<TableHandle>,
+    tables: BTreeMap<TableHandle, Table>,
+}
+
+/// A structure representing a single table.
+struct Table {
+    handle: TableHandle,
+    key_layout: MoveTypeLayout,
+    value_layout: MoveTypeLayout,
+    content: BTreeMap<Vec<u8>, GlobalValue>,
+    size_delta: i64, // The sum of added and removed entries
+}
+
+/// The field index of the `handle` field in the `Table` Move struct.
+const HANDLE_FIELD_INDEX: usize = 0;
+
+// =========================================================================================
+// Implementation of Native Table Context
+
+impl<'a> NativeTableContext<'a> {
+    /// Create a new instance of a native table context. This must be passed in via an
+    /// extension into VM session functions.
+    pub fn new(txn_hash: u128, resolver: &'a dyn TableResolver) -> Self {
+        Self {
+            resolver,
+            txn_hash,
+            table_data: Default::default(),
+        }
+    }
+
+    /// Computes the change set from a NativeTableContext.
+    pub fn into_change_set(self) -> PartialVMResult<TableChangeSet> {
+        let NativeTableContext { table_data, .. } = self;
+        let TableData {
+            new_tables,
+            removed_tables,
+            tables,
+        } = table_data.into_inner();
+        let mut changes = BTreeMap::new();
+        for (handle, table) in tables {
+            let Table {
+                value_layout,
+                content,
+                ..
+            } = table;
+            let mut entries = BTreeMap::new();
+            for (key, gv) in content {
+                match gv.into_effect()? {
+                    GlobalValueEffect::Deleted => {
+                        entries.insert(key, None);
+                    }
+                    GlobalValueEffect::Changed(new_val) => {
+                        let new_bytes = serialize(&value_layout, &new_val)?;
+                        entries.insert(key, Some(new_bytes));
+                    }
+                    _ => {}
+                }
+            }
+            if !entries.is_empty() {
+                changes.insert(handle, TableChange { entries });
+            }
+        }
+        Ok(TableChangeSet {
+            new_tables,
+            removed_tables,
+            changes,
+        })
+    }
+}
+
+impl TableData {
+    /// Gets or creates a new table in the TableData. This initializes information about
+    /// the table, like the type layout for keys and values.
+    fn get_or_create_table(
+        &mut self,
+        context: &NativeContext,
+        handle: TableHandle,
+        key_ty: &Type,
+        value_ty: &Type,
+    ) -> PartialVMResult<&mut Table> {
+        if let Entry::Vacant(e) = self.tables.entry(handle) {
+            let key_layout = get_type_layout(context, key_ty)?;
+            let value_layout = get_type_layout(context, value_ty)?;
+            let table = Table {
+                handle,
+                key_layout,
+                value_layout,
+                size_delta: 0,
+                content: Default::default(),
+            };
+            e.insert(table);
+        }
+        Ok(self.tables.get_mut(&handle).unwrap())
+    }
+}
+
+impl Table {
+    /// Inserts a value into a table.
+    fn insert(
+        &mut self,
+        context: &NativeTableContext,
+        key: &Value,
+        val: Value,
+    ) -> PartialVMResult<(usize, usize)> {
+        let (gv_opt, _, _) = self.global_value(context, key)?;
+        if gv_opt.is_some() {
+            return Err(partial_abort_error(
+                "table entry already occupied",
+                *ALREADY_EXISTS,
+            ));
+        }
+        let key_bytes = serialize(&self.key_layout, key)?;
+        let key_size = key_bytes.len();
+        // Need to serialize for cost computation
+        let val_size = serialize(&self.value_layout, &val)?.len();
+        self.content
+            .entry(key_bytes)
+            .or_insert_with(GlobalValue::none)
+            .move_to(val)?;
+        self.size_delta += 1;
+        Ok((key_size, val_size))
+    }
+
+    /// Borrows a reference to a table (mutable or immutable).
+    fn borrow_global(
+        &mut self,
+        context: &NativeTableContext,
+        key: &Value,
+    ) -> PartialVMResult<(Value, usize, usize)> {
+        let (gv_opt, key_size, val_size) = self.global_value(context, key)?;
+        let gv = gv_opt.ok_or_else(|| partial_abort_error("undefined table entry", *NOT_FOUND))?;
+        let val = gv.borrow_global()?;
+        Ok((val, key_size, val_size))
+    }
+
+    /// Removes an entry from a table.
+    fn remove(
+        &mut self,
+        context: &NativeTableContext,
+        key: &Value,
+    ) -> PartialVMResult<(Value, usize, usize)> {
+        let (gv_opt, key_size, val_size) = self.global_value(context, key)?;
+        let gv = gv_opt.ok_or_else(|| partial_abort_error("undefined table entry", *NOT_FOUND))?;
+        let val = gv.move_from()?;
+        self.size_delta -= 1;
+        Ok((val, key_size, val_size))
+    }
+
+    /// Checks whether a key is in the table.
+    fn contains(
+        &mut self,
+        context: &NativeTableContext,
+        key: &Value,
+    ) -> PartialVMResult<(Value, usize, usize)> {
+        let (gv_opt, key_size, val_size) = self.global_value(context, key)?;
+        let val = Value::bool(gv_opt.and_then(|v| v.exists().ok()).unwrap_or(false));
+        Ok((val, key_size, val_size))
+    }
+
+    /// Compute the size of a table.
+    fn length(&mut self, context: &NativeTableContext) -> PartialVMResult<(u64, usize, usize)> {
+        let remote_size = context
+            .resolver
+            .table_size(&self.handle)
+            .map_err(|err| partial_extension_error(format!("remote table size failed: {}", err)))?;
+        let effective_size = (remote_size as i128) + (self.size_delta as i128);
+        if effective_size < 0 {
+            Err(partial_extension_error("inconsistent table size"))
+        } else {
+            Ok((effective_size as u64, 0, 0))
+        }
+    }
+
+    /// Destroys a table.
+    fn destroy_empty(&mut self, context: &NativeTableContext) -> PartialVMResult<(usize, usize)> {
+        let (len, _, _) = self.length(context)?;
+        if len > 0 {
+            Err(partial_abort_error(
+                "table is not empty and cannot be destroyed",
+                *NOT_EMPTY,
+            ))
+        } else {
+            Ok((0, 0))
+        }
+    }
+
+    /// Gets the global value of an entry in the table. Attempts to retrieve a value from
+    /// the resolver if needed. Aborts if the value does not exists. Also returns the size
+    /// of the key and value (if a value needs to be fetched from remote) for cost computation.
+    fn global_value(
+        &mut self,
+        context: &NativeTableContext,
+        key: &Value,
+    ) -> PartialVMResult<(Option<&mut GlobalValue>, usize, usize)> {
+        let key_bytes = serialize(&self.key_layout, key)?;
+        let key_size = key_bytes.len();
+        let mut val_size = 0;
+        if !self.content.contains_key(&key_bytes) {
+            // Try to retrieve a value from the remote resolver.
+            match context
+                .resolver
+                .resolve_table_entry(&self.handle, &key_bytes)
+                .map_err(|err| {
+                    partial_extension_error(format!("remote table resolver failure: {}", err))
+                })? {
+                Some(val_bytes) => {
+                    val_size = val_bytes.len();
+                    let val = deserialize(&self.value_layout, &val_bytes)?;
+                    self.content
+                        .entry(key_bytes.clone())
+                        .or_insert(GlobalValue::cached(val)?);
+                }
+                None => return Ok((None, key_size, val_size)),
+            }
+        }
+        Ok((
+            Some(self.content.get_mut(&key_bytes).unwrap()),
+            key_size,
+            val_size,
+        ))
+    }
+}
+
+// =========================================================================================
+// Native Function Implementations
+
+/// Returns all natives for tables.
+pub fn table_natives(table_addr: AccountAddress) -> NativeFunctionTable {
+    native_functions::make_table(
+        table_addr,
+        &[
+            ("Table", "new_table_handle", native_new_table_handle),
+            ("Table", "add_box", native_add_box),
+            ("Table", "length_box", native_length_box),
+            ("Table", "borrow_box", native_borrow_box),
+            ("Table", "borrow_box_mut", native_borrow_box),
+            ("Table", "remove_box", native_remove_box),
+            ("Table", "contains_box", native_contains_box),
+            ("Table", "destroy_empty_box", native_destroy_empty_box),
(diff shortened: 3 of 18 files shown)


language/extensions/move-table-extension/sources/Table.move
```diff
@@ -0,0 +1,44 @@
+/// Type of large-scale storage tables.
+module Extensions::Table {
+    struct Table<phantom K, phantom V> has store {
+        handle: u128
```
davidiw, line 4: Could we provide a count of contents so that modules can delete them if they are empty and would like to do so? My exploratory nft work makes use of this concept.

language/extensions/move-table-extension/sources/Table.move
```diff
@@ -0,0 +1,44 @@
...
+    /// Remove from `table` and return the value which `key` maps to.
+    /// Aborts if there is no entry for `key`.
+    native public fun remove<K, V>(table: &mut Table<K, V>, key: &K): (K, V);
+
+    /// Returns true iff `table` contains an entry for `key`.
+    native public fun contains<K, V>(table: &Table<K, V>, key: &K): bool;
+
+    /// Insert the pair (`key`, `val`) to `table`.
+    /// Aborts if there is already an entry for `key`.
+    native public fun insert<K, V>(table: &mut Table<K, V>, key: K, val: V);
```
davidiw, line 43: Yeah, this is the right design for insert. Many libraries return back the value if it exists. I'm smart contract development, I found that there's likely different behavior if there are two values at the same key, e.g. merge. If we return the value in move, we then need to deal with dropping option none or other nonsense in the common case, which isn't ergonomic.

davidiw (commented):
Quick first pass, but I like it!

language/extensions/move-table-extension/sources/Table.move
```diff
@@ -0,0 +1,44 @@
...
+
+    /// Create a new Table.
+    public fun new<K, V>(): Table<K, V> {
+        Table{handle: new_table_handle()}
+    }
+
+    native fun new_table_handle(): u128;
+
+    /// Destroy a table. The table must be empty to succeed.
+    native fun destroy(table: Table);
```
tnowacki, line 15: Maybe destroy_empty? I think we used that naming in Vector, but I'd have to double check. Whatever that name is, let's choose the name that is consistent with Vector's empty resource case

language/extensions/move-table-extension/sources/Table.move
tnowacki: +1 exposing a count might be helpful to know when it is safe to call `destroy`

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -0,0 +1,534 @@
...
+    Value::simple_deserialize(bytes, layout)
+        .ok_or_else(|| partial_extension_error("cannot deserialize table key or value"))
+}
+
+fn partial_extension_error(msg: impl ToString) -> PartialVMError {
+    PartialVMError::new(StatusCode::VM_EXTENSION_ERROR).with_message(msg.to_string())
+}
+
+fn partial_abort_error(msg: impl ToString) -> PartialVMError {
+    PartialVMError::new(StatusCode::ABORTED).with_message(msg.to_string())
```
tnowacki, line 527: I believe ABORTED expects a substatus? Not sure if we actually verify it anywhere but IIRC, this status code is for the `Abort` operation, and some spots might freak out (at least in debug mode) without it

language/extensions/move-table-extension/Move.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "MoveTableExtension"
+version = "1.0.0"
+
+[addresses]
+Std = "0x1"
+Extensions = "0x2"
```
tnowacki, line 7: Should these be in `[dev-addresses]`? Or is it required for now that it both are at these specific addresses?

language/extensions/move-table-extension/sources/Table.move
wrwg: Done

language/extensions/move-table-extension/src/lib.rs
wrwg: Done, and yes, the substatus is required.

language/extensions/move-table-extension/Move.toml
wrwg: Done.

wrwg:
This PR is now ready and tested, thanks to an exhaustive test suite for tables @junkil-park created for the EVM project.

language/extensions/move-table-extension/sources/Table.move
```diff
@@ -0,0 +1,93 @@
+/// Type of large-scale storage tables.
+module Extensions::Table {
+    /// Type of tables
+    struct Table<phantom K, phantom V> has store {
```
sblackshear, line 4: When I saw this, I initially had some concern that we might need `K: copy` to preserve resource safety. After reading through the current implementation, I am convinced that it is safe. However: 

1. I am concerned that someone might add an API that returns a `K` in the future (e.g., `remove_entry<K, V>(table: &mut Table<K,V>, k: &K): (K, V)` without realizing that this compromises resource safety.

2. I am concerned that this might appear to violate resource safety from the perspective of an outside observer. For example: if you do
```
add_nft_to_table(nft: &NFT, v: u64, table: &mut Table<NFT, u64>) {
  Table::insert(nft, v)
}
```

an an outside entity (e.g., a block explorer) displays the keys of `table` as structured types, `NFT` (and its unique ID) will appear to live in two places at once: in the account of the owner, and in the domain of the table. That is potentially confusing/alarming.

This feels somewhat similar in flavor to Eth attacks like [sleepminting](https://timdaub.github.io/2021/04/22/nft-sleepminting-beeple-provenance/) that aren't smart contract exploits, but can be used to scam clients. The equivalent attack for tables might be something like fooling a client into thinking you sent them an NFT because it's a key of the table the client controls.

The easy way to address both of these would be to require `K: copy`. Does that eliminate any of the use-cases Meta/Aptos care about, and are the use-cases important enough that we're willing to risk (2) (less worried about (1), which can be mitigated through thorough comments + testing)?

wrwg: As regards (1) if K has no copy, would it even be possible to call this API? It really depends on how NFTs are created. If they are truly unique than also two keys which are equal cannot exist at the same time, hence you have no way to call this API (either the NFT is in the table or not). (Clearly, in this case we also need to have insert(k:K,..) instead of insert(k:&K,..). I understand that any native API can break this via serialization but seems to be in the nature of such APIs and we need to carefully review them before we let them in.)

Regards (2), I'm not sure whether I fully follow, but a resource explorer for Move should not display the keys, or if so, should display them clearly marked as references.  Would that solve the issue?

I feel a bit reluctant at this time to restrict the keys, as it seems to cause no technical problem to even have tables in keys, And the Move Prover has no issue with such general keys as long as there is no way to discover them from the table. Also, no-copy NFTs as keys seems to be a use case we need, perhaps @junkil-park can say more about this.

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -0,0 +1,606 @@
...
+
+    // Take the transaction hash provided by the environment, combine it with the # of tables
+    // produced so far, sha256 this and select 16 bytes from the result. Given the txn hash
+    // is unique, this should create a unique and deterministic global id.
+    let mut digest = Sha3_256::new();
+    Digest::update(&mut digest, table_context.txn_hash.to_be_bytes());
+    Digest::update(&mut digest, table_data.new_tables.len().to_be_bytes());
+    let bytes: [u8; 16] = digest.finalize()[0..16].try_into().unwrap();
+    let id = u128::from_be_bytes(bytes);
+    assert!(table_data.new_tables.insert(TableHandle(id)));
```
jolestar, line 377: If multiple tables are created for the same transaction, will the generated ids be duplicated (if they all have the same length)?

wrwg: The adapter needs to provide a unique hash for the transaction (the txn_hash) and table ids would be unique relative to this hash.

jolestar: Got it, it's tables.len(), not table_data.len()

language/extensions/move-table-extension/sources/Table.move
sblackshear: > Regards (2), I'm not sure whether I fully follow, but a resource explorer for Move should not display the keys, or if so, should display them clearly marked as references. Would that solve the issue?

This would solve the issue, but it raises special cases that every client (explorers, wallets, marketplaces, ...) would not have to worry about if not for tables. E.g., "mark it as a reference" isn't something that clients currently have to handle, since on-chain structs can't hold references. To give another example, you can imagine a client implementing an API like `get_all_types<T>(a: address) -> Vec<T>` that returns all top-level or nested values of type `T` in `a`. If the implementation doesn't know to special-case table keys, it will return two copies of `NFT` in code like the example above.

As a language designer who now spends a fair amount of time on clients and the issues Move creates for them, I am coming to appreciate the importance of minimizing client complexity more, particularly because every client has to implement and understand special cases we introduce. I personally weigh this complexity reduction higher than supporting hypothetical use-cases that need non-`copy` keys (particularly if we don't have any in mind), but I am only one opinion here.

> As regards (1) if K has no copy, would it even be possible to call this API?

Good point on--this particular API is indeed impossible to call for a `copy` type.

language/tools/move-unit-test/src/extensions.rs
```diff
@@ -0,0 +1,74 @@
...
+    if let Ok(cs) = cs {
+        if !cs.new_tables.is_empty() {
+            writeln!(
+                w,
+                "new tables {}",
+                cs.new_tables.iter().map(|h| h.to_string()).join(", ")
+            )
+            .unwrap();
+        }
+        if !cs.new_tables.is_empty() {
```
sblackshear, line 56:
```suggestion
        if !cs.removed_tables.is_empty() {
```

language/extensions/move-table-extension/sources/Table.move
wrwg: We could sha256 all keys before we apply them. That is what the EVM does. This basically obfuscates them. I do not know how explorers for EVM chains deal with this but in fact I can't imagine that EVM storage can be meaningfully explored.

sblackshear: That would also work.

language/extensions/move-table-extension/sources/Table.move
msmouse: Supporting length turns some insertions (those creating new entries) and removals writing an additional metadata node in the storage. Do we consider a version of map that doesn't know its length and not safely droppable (and makes its embodying structure not droppable as well)? Because the deletion refund won't matter much, I suspect there won't be anybody who bothers to clean up tables anyway, especially the empty table itself.

And the length-aware version can be implemented on top of that.

msmouse:
(I see the comment was resolved, so reposting)

Supporting length turns some insertions (those creating new entries) and removals writing an additional metadata node in the storage. Do we consider a version of map that doesn't know its length and not safely droppable (and makes its embodying structure not droppable as well)? Because the deletion refund won't matter much, I suspect there won't be anybody who bothers to clean up tables anyway, especially the empty table itself.

And the length-aware version can be implemented on top of that.

language/extensions/move-table-extension/sources/Table.move
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /// Returns true iff `table` contains an entry for `key`.
+    public fun contains<K, V>(table: &Table<K, V>, key: &K): bool {
+        contains_box<K, V, Box<V>>(table, key)
+    }
+
+    #[test_only]
+    /// Testing only: allows to drop a table even if it is not empty.
+    public fun drop_unchecked<K, V>(table: Table<K, V>) {
```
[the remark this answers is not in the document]
tnowacki: I think it will be useful for testing so you can run the code without Move complaining about not consuming the table handle. 

If I follow correctly, this native function is a no-op and `destroy_empty_box` is where the real logic is

language/extensions/move-table-extension/sources/Table.move
tnowacki: As far as the language itself goes (not concerning external observers or expectations), I don't mind either way since there isn't *really* a safety issue here.

But it might be confusing to anyone used to Move or Rust or similar setups that `insert` has the signature `insert(&K, V)`. I think might seem bizarre that you able to insert into a collection without actually giving it the value. Especially considering that if/when we have `Map`s, our `Map::insert` will likely look something like `Map::insert(K, V)` with an owned key as input. Additionally if we don't have `K: copy` for maps, we will have to do this strange key return on `remove`, e.g. `Map::remove(&K, V): Option<(K, V)>` where we pretend we have tuples :p. 

If you really need the `Table::insert(&K, V)` you could always implement it with `BCS::to_bytes`, e.g. 
```
struct Table<K, V> { ... } // no copy on the K/V
native public fun insert<K: copy, V>(table: &Table<K, V>, key: K, value: V);
public fun bcs_insert<K, V>(table: &Table<K, V>, key_ref: &K, value: V) { insert(table, BCS::to_bytes(key_ref), value) }
```

TL;DR while it doesn't really make a difference, but not having `copy` on `K` might lead to some confusing disparities with other Map-like APIs

tnowacki, line 4: Is see you have `V: store` on `new` so we should be fine, but should also have `V: store` on the declaration?

I understand it isn't necessary, but it might be less confusing (that is `struct Table<phantom K, phantom V: store> has store`)

wrwg: > I think might seem bizarre that you able to insert into a collection without actually giving it the value.

I don't think that is "bizarre" given that you aren't able to discover keys from a table. That tables can't have keys discovered is a feature which differentiates them from "maps". People coming from Solidity/EVM smart contracts are used to this, it's probably more common to smart contract users than the Rust style.

For Move it is essential that keys can't be recovered, and also that BCS::to_bytes is not explicitly called by the user. Otherwise the Move prover stops working. The prover has no realistic way to reason about serialization (of symbolic values), or to reason about key ordering if keys could be enumerated. The current API goes just such far as verification is able to support. It's by intention like it is.

wrwg:
> (I see the thread was resolved, so reposting)
> 
> Supporting length turns some insertions (those creating new entries) and removals writing an additional metadata node in the storage. Do we consider a version of map that doesn't know its length and not safely droppable (and makes its embodying structure not droppable as well)? Because the deletion refund won't matter much, I suspect there won't be anybody who bothers to clean up tables anyway, especially the empty table itself.
> 
> And the length-aware version can be implemented on top of that.

If we drop length, we still need the ability to check for emptiness to implement `destory_empty` correctly. I figured there is not a big difference technically to support `is_empty` or `length', both need extra metadata in storage, so I stick to the later also because of @davidiw's ask for it.

language/extensions/move-table-extension/sources/Table.move
jolestar: I agree with @tnowacki comments. Although we can't retrieve its original key from Map/Table, explicitly indicating that the key needs `copy` and letting the developer serialize it in the Move module would allow him to understand how the key is mapped to the value.

However, if we consider the move-prover, I may need to understand the move-prover in more detail about this.

wrwg: > letting the developer serialize it in the Move module would allow him to understand how the key is mapped to the value.

The intention of this API is to *not* let the developers reason about how keys are implemented. For example, whether leys are actually implemented by BCS or by some sha256 (as discussed earlier in this thread) is something this API intentionally hides from the user, thus leaving implementations the choice how to do this. 

As regards users explicitly calling BCS, I wish we would have a linter which disallows this. It's really a bad idea to do so in user contracts. All the arguments @sblackshear brought up in this thread matter for any visible BCS value on the chain, so we should just not allow calling this, apart for some needed runtime internals.

wrwg: PS. Insead of BCS, we should have `sha256<T>`. This would eliminate the need for `sha256(to_bytes(x))`, which is the only practical use case for BCS in user code which comes to my mind.

language/extensions/move-table-extension/sources/Table.move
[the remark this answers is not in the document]
wrwg: It's used in the current testsuite, as Todd describes.

language/extensions/move-table-extension/tests/TableTests.move
```diff
@@ -0,0 +1,204 @@
...
+    #[expected_failure(abort_code = 51155)]
+    fun test_destroy_fail() {
+        let t = T::new<u64, u64>();
+        T::insert(&mut t, &1, 2);
+        assert!(*T::borrow(&t, &1) == 2, 1);
+        T::destroy_empty(t) // expected to fail
+    }
+
+     #[test(s = @0x42)]
+    fun test_primitive(s: signer) acquires S {
```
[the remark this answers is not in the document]
wrwg: Done, and found and fixed a bug.

language/extensions/move-table-extension/sources/Table.move
wrwg: Done.

language/tools/move-unit-test/src/extensions.rs
wrwg: Done.

language/extensions/move-table-extension/sources/Table.move
wrwg: I pushed a commit which renames `Table::insert` to `Table:add`. This seems to be better given the unusual semantics compared to Rust, which is not only `&K`, but also that we do abort on existence. I still think the current signature of the API is the right way to go. These aren't the same as Rust maps.

msmouse:
> If we drop length, we still need the ability to check for emptiness to implement destory_empty correctly. I figured there is not a big difference technically to support is_empty or `length', both need extra metadata in storage, so I stick to the later also because of @davidiw's ask for it.

I mean, in that (default) flavor of table, `destroy_empty()` is not supported either. -- If you don't have a plan to delete your table, use the flavor which saves gas on inserting and deleting items.

language/extensions/move-table-extension/sources/Table.move
jolestar: I'm just worried that developers might misunderstand that value is associated with the ref of the key. But we actually associate the value with the value of the key. Especially if the key is a resource.

In the BCS case, if the chain does not support BCS, it should support other serialize modules, the developer can serialize the resource with other serialize modules.

This feels like a big decision. Maybe I can ask some Move contract developers how they understand this at the meeting tomorrow.

wrwg:
> > If we drop length, we still need the ability to check for emptiness to implement destory_empty correctly. I figured there is not a big difference technically to support is_empty or `length', both need extra metadata in storage, so I stick to the later also because of @davidiw's ask for it.
> 
> I mean, in that (default) flavor of table, `destroy_empty()` is not supported either. -- If you don't have a plan to delete your table, use the flavor which saves gas on inserting and deleting items.

Tables which have different gas models may be confusing for users. Why can't we have one which has the right properties (including refund)? If this is not supported by a given platform now, refund might be ignored for now.

wrwg:
FYI I'm leaving town tomorrow afternoon for ETAPS conference and will likely not be able to work on this PR again before being back 11/4. If we want to land this and possibly iterate on it, we need to do it in the next 24h.

language/extensions/move-table-extension/sources/Table.move
tnowacki: > The intention of this API is to not let the developers reason about how keys are implemented. For example, whether leys are actually implemented by BCS or by some sha256 (as discussed earlier in this thread) is something this API intentionally hides from the user, thus leaving implementations the choice how to do this.

I personally wouldn't get that as a developer. This sort of API would do the opposite and *force me* to think about how keys are implemented, since they are doing something unusual for Move, and taking the key by reference. If it was an owned key value pair, I think it is easy to say "oh, it is putting the key value pair in the map, easy". But with this reference, it raises questions that would force one to think about whats going on under the hood "how is it using the value if it isn't owned? Is it actually storing the reference? Why can't I do that in my own structs?" 

----

Given @jolestar's and @sblackshear's similar concern. Maybe it would be easier to stick to `K: copy + drop` for this initial PR?

You can still do a Sha256 wrapper, or any other wrapper you want (as a separate module I now realize). We can then explore all of these separate wrappers as we learn more about how developers will use Tables. In short, starting with `K: copy + drop` might be better for initial exploration

msmouse:
> Why can't we have one which has the right properties (including refund)?

The refund didn't work for Ethereum (people don't bother to delete, except GasCoin), and they intend to [remove it](https://eips.ethereum.org/EIPS/eip-3298)

My thinking is to avoid arbitraging on gas, the refund must be cheaper than creation, and modern blockchains want the creation to be cheap -- as a result, no one will bother deleting.  So what I really tried to say is, I only want to not-destroyable version. 

And rent / state expiry is the right answer for the state explosion problem, not intentional deletion. It's a separate discussion, I agree we shouldn't block on that today.

language/extensions/move-table-extension/sources/Table.move
tnowacki: > Your's and Jolestar's main argument I can see is that you don't like it because it is not the same as in Rust. 

I don't think thats the argument either of us are making. It at least isn't mine. I mentioned Rust briefly only in that they also have a similar linear type system. 
All of my points are really about how APIs work in Move. Maybe that will help reframe what I have said? 

In essence, I, as a Move developer, wouldn't expect any collection API to behave like this. (And our Map APIs almost certainly won't). I think the disparity will be confusing and force users to think about Tables very differently, which may or may not be a good thing.
Going a bit further just on the storage side of that API, all of our existing interactions with "storage" are over owned values. You can modify by reference, but there is no insertion/removal by reference. I think it might feel confusing that the keys for tables don't behave like other storage values, even though they actually appear in storage. So what I was trying to argue is it feels more straightforward to me to have tables operate over owned keys (with `copy + drop` for API simplicity), and then have wrapper modules if you want to do things like Sha256 of "resoruce" keys by reference.

wrwg: A table is not a regular collection, that's why we call it table and not "map". *A table does not store any keys.* There is no way how you can retrieve keys from a table. Therefore moving or copying keys into a table on insert is logically inconsistent. In Move it is idiomatic to pass a reference if you do not transfer ownership; that is what is happening on insert.

If you move keys into a table on insert, the next question would be: why can't you can get them back from the table? The API would be incomplete and illogical. If you make now the keys copyable, you just hide that inconsistency. Now a copy of the key is conceptually stored in the table, but you still can't get it back. 

The way how it is right now is really the most idiomatic way for Move I can imagine, given all the design constraints we already agree upon. I believe we still need to introduce real 'maps' which can be iterated and are not large scale, but this is a different discussion,

davidiw: As a developer hacking on a pseudo-table,  I definitely state that an owned key or making it copyable is irrelevant unless you want to support enumeration of the table. Considering that is a non-goal, the user of a table would need to concoct some other method to keep track of keys. I mean if a key cannot be generated or enumerated, how on earth would a user retrieve the data from the table? I cannot fathom the value in keeping such a key.

davidiw (commented):
I'd like to get this diff landed before Wolfgang goes on leave tomorrow. I think this is reasonable given that the work is exploratory and not law. The code as is written is agreeable to our requirements from an interface perspective. There hasn't been claims of breaking those requirements. It would give us a chance to actually evaluate the ergonomics and provide meaningful feedback.

Are there any objections to landing this and continuing the conversation after one or more parties begin to build upon this?

If there aren't any objections, I'll swing by tomorrow at 9 AM PST.

language/extensions/move-table-extension/sources/Table.move
tnowacki: > I definitely state that an owned key or making it copyable is irrelevant unless you want to support enumeration of the table

`copy + drop` doesn't have to do with enumeration, so I'm not sure I see where you are coming from.

I think the concerns @jolestar and I both share are that 
- the API feels cohesive with other libraries.
- the API feels logical in conveying intended behavior to a user of the module
Did I express your concerns correctly, @jolestar?

> A table does not store any keys. There is no way how you can retrieve keys from a table. Therefore moving or copying keys into a table on insert is logically inconsistent. 

I think if we can nail this or make it apparent to the user, that would be great. Maybe even having the reference in the API will help a developer, who is comparing this to Map or other libraries, stop and say "what is going on here? How is this different?" and then that will help them understand what is going on in the Table in general.

tnowacki (approved):
I think we should still discuss the API at tomorrows meeting; I want to make sure @jolestar's concerns are fully understood. 
But we can make those changes after this lands.

jolestar:
> I think we should still discuss the API at tomorrows meeting; I want to make sure @jolestar's concerns are fully understood. But we can make those changes after this lands.

I agree. Maybe we should first add the `copy` restriction to the key, and then remove the `copy` restriction if there is enough practical requirement to use the resource as the key. 

Removing `copy` will not cause compatibility problems, but if we want to add the `copy` restriction, it will cause compatibility problems.

wrwg:
> Removing `copy` will not cause compatibility problems, but if we want to add the `copy` restriction, it will cause compatibility problems.

I think there is a misunderstanding of the issue we are talking about, @jolestar . Currently its `key: &T`, and it doesn't matter what abilities `T` has. If we make it `key: T`, it's quite a different API and only then comes the `copy` into the picture. 

> copy + drop doesn't have to do with enumeration, so I'm not sure I see where you are coming from.

I think they do, @tnowacki , in the sense that you can only enumerate keys from a table if the table actually owns them. To own them, you pass in `key: T` (doesn't really matter now whether copy or not), If you don't own, the idiomatic style in Move is `key: &T`. The table currently does not own the keys, that is what makes it different from maps.

wrwg:
Thanks @tnowacki for approving, and we definitely should discuss and refine this further (this and next week's mtg however I cannot attend).

Also we should discuss what @msmouse brought up, whether we remove the destroy_empty from the API, and make tables 'eternal'. I can see that this makes live easier for the storage backend, and the are probably rare cases where users want to destroy a table -- however, users might be quite surprised *if* they run into such cases and can't destroy a table (and by extension, any resource which aggregates a table).

/land

wrwg:
Need a restamp because of merge conflict. @davidiw @msmouse @tnowacki

msmouse (approved):
Thank you @wrwg

jolestar:
> think there is a misunderstanding of the issue we are talking about, @jolestar . Currently its `key: &T`, and it doesn't matter what abilities `T` has. If we make it `key: T`, it's quite a different API and only then comes the `copy` into the picture.

I prefer  `K: copy + drop` and `key: K`. 

`K: copy + drop` and `&key: K` are also acceptable. 

For a no copy example:

```rust
module MyModule{

  struct Foo {
     f1: u64,
  }

  public fun table_test(){
     let table<Foo,u64> = //get a table;
     let foo = // get a Foo instance
     table.insert(&foo,1);
     //update the key
     foo.f1 = 2;
     //how the developer this about this?
     let v = table.borrow(&foo); 
  }
}
```

For a copy example:

```rust
module MyModule{
  struct Foo has copy,drop{
     f1: u64,
  }

  public fun table_test(){
     let table<Foo,u64> = //get a table;
     let foo = // get a Foo instance
     table.insert(copy foo,1);
     //update the key
     foo.f1 = 2;
     //It is more easy for developers to understand 
     //that the value is associated with the key of the copy, 
     //so v can not get via a modified key.
     let v = table.borrow(&foo); 
  }
}
```

wrwg:
> > think there is a misunderstanding of the issue we are talking about, @jolestar . Currently its `key: &T`, and it doesn't matter what abilities `T` has. If we make it `key: T`, it's quite a different API and only then comes the `copy` into the picture.
> 
> I prefer `K: copy + drop` and `key: K`.
> 
> `K: copy + drop` and `&key: K` are also acceptable.
> 
> For a no copy example:
> 
> ```rust
> module MyModule{
> 
>   struct Foo {
>      f1: u64,
>   }
> 
>   public fun table_test(){
>      let table<Foo,u64> = //get a table;
>      let foo = // get a Foo instance
>      table.insert(&foo,1);
>      //update the key
>      foo.f1 = 2;
>      //how the developer think about this?
>      let v = table.borrow(&foo); 
>   }
> }
> ```
> 
> For a copy example:
> 
> ```rust
> module MyModule{
>   struct Foo has copy,drop{
>      f1: u64,
>   }
> 
>   public fun table_test(){
>      let table<Foo,u64> = //get a table;
>      let foo = // get a Foo instance
>      table.insert(copy foo,1);
>      //update the key
>      foo.f1 = 2;
>      //It is more easy for developers to understand 
>      //that the value is associated with the key of the copy, 
>      //so v can not get via a modified key.
>      let v = table.borrow(foo); 
>   }
> }
> ```

Thanks Jolestar, this is an Interesting example. However, a developer that would make the mistake in the first example would also have a missing understanding of fundamentals of reference semantics in Move.  The value behind a `&foo` can never be mutated in Move (or Rust): this is the essence of borrow semantics. I'm not sure whether such a developer who would be able to successfully use Move at all, however supposedly simple we define the table API. Just a thought. We need to understand our developers better.

jolestar:
> However, a developer that would make the mistake in the first example would also have a missing understanding of fundamentals of reference semantics in Move. The value behind a `&foo` can never be mutated in Move (or Rust): this is the essence of borrow semantics.

Although this code is a bit stupid(😂) and modifying the key is not a good idea in any language, it may happen and the above example code works, especially if the update key operation does not happen in the same method.

wrwg:
Even though I do think `key: &T` is the right approach for tables, I want to outline what would be the compromise I could agree with if the majority of us continues to feel uncomfortable with this.  

- There will be a new type `Hash` which can be implemented as a wrapper around `u256`. That type is fully abstract and only supports equality.
- There will be a function `fun sha256<T>(x: &T): Hash` part of the standard library. That (and perhaps some other hash algorithms) is the only way one can create an instance of `Hash`. 
- Now if someone wants to use non-copyable values as keys, they would have to use `table.insert(sha256(k), v)`.

This design will allow the Move Prover to reason about tables. Because of the opaqueness of `Hash`, the prover can choose any internal representation for it as long as it satisfies the basic properties of perfect hashes (`x == y <=> sha256(x) == sha256(y)` (including just a wrapper around the original value, which is likely the best for the prover).

This should address @tnowacki and @jolestar's concerns, and also @sblackshear concern (which I think is a different one), plus it also works with the requirements of the prover (@DavidLDill @shazqadeer).

I still believe passing `key: &T` is more idiomatic for Move, but the above is something I could live with as well.

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -0,0 +1,606 @@
...
+    Remove,
+    Contains,
+}
+
+/// The native table context extension. This needs to be attached to the NativeContextExtensions
+/// value which is passed into session functions, so its accessible from natives of this
+/// extension.
+pub struct NativeTableContext<'a> {
+    resolver: &'a dyn TableResolver,
+    txn_hash: u128,
```
msmouse, line 99: Shall we worry about collisions here? Shall we do `[u8; 32]`? 

cc @tnowacki @wrwg @davidiw @alinush

davidiw: Yeah, it would be great to have a stronger hash than just 16-bytes. We moved Aptos away from 16-byte addresses due to the weak properties and lack of user ergonomics as a result.

wrwg: That is an interesting question. I think we need to consult a cryptography expert here (@kchalkias?). My thought was that while we can hash the known data on the planet in u256, we don't need that precision in this case, because the variety of the possible values is restricted.

If we do 256, we should use the type U256 from the primitive_types crate instead of the direct array.

wrwg: ... and to note: it seems common in the Ethereum world for such cases to compute 256 bit hashes (keccak in their world) but then slice them down to 8 or 16 bytes. This is theoretically sound because any bit in the hash has the same randomness. Indeed the collision probability may rise.

kchalkias: There are major UX benefits with 128bit addresses (many wallet users have noticed that anyway re usability), and the original decision was that unlike Ethereum, by design we restricted sending tx to future addresses (to avoid dealing with offchain attacks). However, global demand for these applications have raised since then, and the selection is actually dependent on whether on each address creation we check our hashset of existing addresses.
So today we need to increase the length, to err on the safe side.

The reality is many cool ideas would benefit i.e., from 160bit addresses (half of world's chains use this size anyway), a) due to compatibility reasons, b) fitting a transactions in an sms, c) improve caching requirements, d) tx network payload size etc.

But, we recently realized that Ambush Attacks affect Ethereum's 160bit addresses too, in some non-updated node libraries. All in all, in 2022 (I would suggest this to other chains too, including to our Ethereum friends)  - we should gradually move to 32 byte addresses as @davidiw mentioned.
Attaching Ambush attack presentation from late 2020 [Ambush Attacks on 128 and 160 bit addresses - Kostas Chalkias](https://drive.google.com/file/d/1llM7RYxarlCQl7VX4UmHAczSXZ_83WaX/view?usp=sharing)

wrwg: Thanks for the explanations. Yes I agree for addresses, and aptos already pushed for extending Move addresses to 256 bits.

Here the situation is a bit different. We want create a unique key for an object in storage (like a 'guid'). The way how we do that is take a base hash identifying the transaction and combine with a counter local to that transaction. Currently we use a 128 bit base hash, combine this with the 16 bit counter, hash the result into 256 bits, then truncate the result to 128 bits to obtain the key. The question is whether this should follow the same scrutiny as for addresses, i.e. do all in the 256 bits. May lay man's guts here were telling me that the information we are hashing is not that diverse than arbitrary addresses, but I might get that wrong, and the domain you are hashing over doesn't really matter here. For the record, EVM does something similar like this e.g. in the domain of method selectors.

kchalkias: The ambush attack I described above probably applies to objects as well, right? If we eventually truncate to 128bit objectIDs, it is theoretically possible, even with 256bits addresses, for one to off-chain compute 2 256bit addresses where objectID1 = Hash(address1, counter1) = objectID2 = Hash(address2, counter2) for some counter1 and counter2. So, if when adding new objects we don't check against `HashSetofObjects.exists(ObjectID)`, then one can overwrite the identifier of a previous objectID.

How can this be exploited without a global hashSet check each time? I guess an example would be
handcraft the ambush attack, in such a way where instead of accessing Object1 data, you confuse the validators to take into account Object2 data during computations, and somehow drain some account?

jolestar:
> * Now if someone wants to use non-copyable values as keys, they would have to use `table.insert(sha256(k), v)`.

Is the table's insert function will be `public fun insert<V>(&mut Table<V>, key: Hash, v: V)`?

language/extensions/move-table-extension/src/lib.rs
alinush: Collisions on 128-bit hashes happen w/ probability 1/2^{64}. This is a *high* probability. So, if the price to pay for a collision is just a performance hit, of course that's fine. But if the price is a break in the correctness of the execution (because the colliding key's value is maliciously overwritten) then use 256-bit hashes.

In general, when you are truncating cryptographic hashes, **you are playing with fire**. For example, truncating hashes by getting the last n/2 bits can be completely *insecure* (see https://crypto.stackexchange.com/questions/3153/sha-256-vs-any-256-bits-of-sha-512-which-is-more-secure/3155#3155).

While we know that truncating SHA2 hashes is not susceptible to these attacks, if in the future the hash function changes, there could be problems.

kchalkias: The main question is there exist applications that rely on pre-image security only. For instance Flow (Dapper Labs) uses 64bit addresses, because they are assigned by the network. In theory you can have a system where each account is being created by a counter of insertion order. That was the original plan for Libra too and the only reason 128bit was selected was to avoid land grabbing. So if you always check if an address already exists at creation and your application logic is not affected by general collisions, but only on pre-image, 2nd preimage (sometimes multi-target preimage too), then many systems tolerate truncated hashes (incl Bitcoin and Eth).
Also note that in Diem there is decoupling between the address and the spending key. The spending key hash (auth_key) is not truncated, only the address id is, and that's why it could even be a counter (assuming we tolerate long range attacks).

wrwg:
> > * Now if someone wants to use non-copyable values as keys, they would have to use `table.insert(sha256(k), v)`.
> 
> So， the table's insert function will be `public fun insert<K: copy, V>(&mut Table<K, V>, key: K, v: V)`? I'm glad to see this.

I have not committed to this yet, I'm just saying it could be a compromise if there are better arguments for why not `key: &K` ;-) I have not seen those yet, sorry. So keep them coming please.

Also, if we settle on this, it is essential that the other conditions are met which I listed, which keep Move verifiable and analyzable.

davidiw:
> 
> I have not committed to this yet, I'm just saying it could be a compromise if there are better arguments for why not `key: &K` ;-) I have not seen those yet, sorry. So keep them coming please.
> 
> Also, if we settle on this, it is essential that the other conditions are met which I listed, which keep Move verifiable and analyzable.

Unless we're committed to storing the value of Key in the table, this seems to give the false impression. Also what are the performance considerations here? The key may be an arbitrarily large value and a copy may have non-zero gas associated with it.

To counter @jolestar's comment, I can see the confused programmer moving a key value into the key and then wondering how do they get their key back. I think the fact that this is a table and not a map implies a different set of expectations. Perhaps the right solution is to expose a Map as well and then let the VM integrators decide if they want to support both or either or. Mind you, a Map can theoretically be stored as a few tables -- Table<&Key, Index>, Table<&Index, Key>, Table<&Key, Value> -- then you get the ability to count, enumerate, and get keys back.

PR #151, opened 2022-03-26
[move-vm] Revisit session extension API.

wrwg:
This puts the extensions into the Session struct instead of passing them as parameters to exec functions. This design was in some earlier version not possible, as the execute functions had extra type parameters for extensions; however, since we use the new `NativeContextExtensions` object, its natural and cleaner.

One does now

```
let session = vm.new_session_with_extensions(remote, extensions);
<<work with the session>>
let (change_set, events, extension) = session.finish_with_extensions()?;
```

Furthermore the `ExecutionResult::Success` has been extended to have a new field for extensions, which is a breaking change, but a marginal one.

In this design, if one wants to access the extensions on a failed execution, we can just add another function to the Session API to do so. This is currently not there but can be easily added.

## Motivation

Better API.

Yes

### DIFF
--- a/language/move-vm/runtime/src/move_vm.rs
+++ b/language/move-vm/runtime/src/move_vm.rs
@@ -2,7 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    data_cache::TransactionDataCache, native_functions::NativeFunction, runtime::VMRuntime,
+    data_cache::TransactionDataCache,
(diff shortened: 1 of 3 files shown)


sblackshear (dismissed):
LGTM.

 The minor breaking change reminds me of something that I will put on the agenda for the next Move meeting: a plan to do a release (it's been awhile!) + publishing to crates.io. All Move customers are currently depending on Move via direct git deps w/ a hash.

Issue #147, opened 2022-03-25
[Bug] Move Prover proves wrong spec for shifting operator


whst:
# 🐛 Bug

Move prover proves  wrong spec for shifting operator.

## To reproduce

Move prover can successfully find the verification error for the following module:
```rust
module 0x1234::Test {
    fun shift(x: u64): u64 { x << 8 }
    spec shift { ensures result == x * 256 + 1; }  // The +1 should cause an error
}
```
But it cannot find the problem for this almost identical version:
```rust
module 0x1234::Test {
    fun shift(x: u64): u64 { x << 7 }
    spec shift { ensures result == x * 128 + 1; }  // The +1 should cause an error
}
```

## Expected Behavior

MVP should report "post-condition does not hold" for both version of the code.

## Additional context

After some investigation, I found that in `move-prover/boggie-backend/src/prelude/prelude.bpl`, `$shl` and `$shr` are defined to support shifting by 8, 16, 32, and 64-bits only. Is that feature on purporse or a bug?

DavidLDill:
I've submitted a PR for this bug, which I hope to be able to land very soon.

Do you expect to be using the Move Prover more in the near future? Are you affiliated with one of the blockchain projects using Move?

Thanks,
Dave
________________________________
From: 五好蒜苔WHsT ***@***.***>
Sent: Thursday, March 24, 2022 5:55 PM
To: diem/move ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [diem/move] [Bug] Move Prover proves wrong spec for shifting operator (Issue #147)


🐛 Bug

Move prover proves wrong spec for shifting operator.

To reproduce

Move prover can successfully find that the verification error for the following module:

module 0x1234::Test {

    fun shift(x: u64): u64 { x << 256 }

    spec shift { ensures result == x * 256 + 1; }  // The +1 should cause an error

}

But it cannot find the problem for this almost identical version:

module 0x1234::Test {

    fun shift(x: u64): u64 { x << 7 }

    spec shift { ensures result == x * 128 + 1; }  // The +1 should cause an error

}

Expected Behavior

MVP should report "post-condition does not hold" for both version of the code.

Additional context

After some investigation, I found that in move-prover/boggie-backend/src/prelude/prelude.bpl, $shl and $shr are defined to support shifting by 8, 16, 32, and 64-bits only. Is that feature on purporse or a bug?

whst:
Yes, I am trying to prove the correctness of the stdlib in starcoin project.

[...]

PR #155, opened 2022-03-28
[move-prover] Fixed bugs in shift

DavidLDill:
Issue #147 reported a bug in the left shift operation where the prover
failed to report an obviously violated postcondition.  The report also
observed that the shift operators were only for a few shift amounts.
It is unfortunate that the unimplemented shift amounts did not result
in a false error report.  Upon inspection of the code, I also found a
"TODO" that shift left doesn't delete bits that are shifted left of
the end of the destination.  In the prover, this would result in cases
where the unbounded integer value would exceed the upper bound imposed
by the bit width of the destination.

The use of a recursive function raises a risk of non-termination, but
it seems to work fine in test cases, including a variable shift
(somewhat to my surprise).

I made the following changes and improvements
1. In prelude.bpl, implemented a recursive function $pow
   to raise an integer value an integer exponent. It explicitly
   returns an undefined result when the exponent is negative,
   to eliminate an obvious possibility for non-termination.
2. Change $shl and $shr to call this function to raise 2 to
   the shift amount, the multiply or divide the shifted value by
   that amount.
3. Added sized shift-left instructions $ShlU8/64/128 which call
   $shl and then mod the result by the appropriate amount to delete
   bits that would be shifted beyond the limits of the destination.
   No such change was required for $Shr, so there is only one $Shr
   operator.
4. Modified bytecode_translator to generation $ShlU8/64/128 instead of
   $Shl.
5. Added function tests for shift operators in shift.move

## Motivation

(Write your motivation for proposed changes here.)

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/language/move-prover/boogie-backend/src/bytecode_translator.rs
+++ b/language/move-prover/boogie-backend/src/bytecode_translator.rs
@@ -1306,10 +1306,17 @@ impl<'env> FunctionTranslator<'env> {
                         let dest = dests[0];
                         let op1 = srcs[0];
                         let op2 = srcs[1];
+                        let sh_type = match &self.get_local_type(dest) {
+                            Type::Primitive(PrimitiveType::U8) => "U8",
+                            Type::Primitive(PrimitiveType::U64) => "U64",
+                            Type::Primitive(PrimitiveType::U128) => "U128",
+                            _ => unreachable!(),
+                        };
                         emitln!(
                             writer,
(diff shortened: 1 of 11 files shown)


wrwg (dismissed):
Great also to see the added tests!

wrwg (approved):

wrwg (approved):

PR #156, opened 2022-03-28
[move-package] Ensure dev-addresses are unique

tnowacki:
## Motivation

- With the recent change (#145) to named addresses being transparent, it is now possible for code to not work for every possible instantiation of addresses
- This fixes that by ensuring unassigned addresses, set in dev addresses, are assigned to a unique address
- We still need to ban/warn against non-named address qualification of module members to fully ensure packages compile properly with any mapping of named addresses 

## Test Plan

new tests 

## Related PRs

- Followup to #145

### DIFF
--- a/language/tools/move-cli/tests/move_unit_tests/assign_dev_addr_for_dep/Move.toml
+++ b/language/tools/move-cli/tests/move_unit_tests/assign_dev_addr_for_dep/Move.toml
@@ -8,7 +8,7 @@ A = "_"
 [dev-addresses]
 Std = "0x1"
 A = "0x2"
-B = "0x2"
+B = "0x3"
 
 [dev-dependencies]
 MoveStdlib = { local = "../../../../../move-stdlib" }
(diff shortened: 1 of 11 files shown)


tnowacki:
@vgao1996 I had an issue where the CLI silenced the output from the package manager in EVM mode. 

For `language/tools/move-cli/tests/move_unit_tests/assign_dev_addr_for_dep/Move.toml`
If you have `B = "0x2"`

I would expect my new error from the package manager to show up, instead I get:

```
test run_all::assign_dev_addr_for_dep/args.evm.txt ... FAILED
Error: Expected output differs from actual output:
Command `package test --evm`:
BUILDING MoveStdlib
BUILDING Bar
BUILDING Foo
Running Move unit tests
[ PASS    ] 0x2::M::nop
Test result: OK. Total tests: 1; passed: 1; failed: 0
Error: Unable to resolve packages for package 'Foo'

Caused by:
    0: While resolving dependency 'Bar' in package 'Foo'
    1: Unable to resolve package dependency 'Bar'
    2: While resolving dependency 'MoveStdlib' in package 'Bar'
    3: While processing dependency 'MoveStdlib'
    4: Unable to find package manifest for 'MoveStdlib' at "Move.toml/./dep/../../../../../../move-stdlib"
```

Should I file an issue?

sblackshear (dismissed):

PR #160, opened 2022-03-29
[move-vm] Removed most RCs from runtime

tnowacki:
## Motivation

- RCs are no longer necessary due to Move's reference safety checks
- Some RCs are still needed to update clean/dirty status or hold onto external memory safely
- I feel this greatly simplifies the layout of values in value_impl.rs
- It does however make `write_ref` unsafe, as it could lead to memory issues if the reference safety checks are wrong or not run
- I do not have a reference heavy bench mark to test if this makes any perf difference. My guess is it won't, but might help with memory usage

## Test Plan

- ran tests
- I'm a bit nervous that our tests might not be exhaustive enough though....


language/move-vm/types/src/values/values_impl.rs
```diff
@@ -607,46 +568,24 @@ impl Value {
...
-            Self::ContainerRef(r) => r.read_ref(),
-            Self::IndexedRef(r) => r.read_ref(),
-        }
-    }
-}
-
-impl StructRef {
-    pub fn read_ref(self) -> PartialVMResult<Value> {
-        self.0.read_ref()
+    fn read_ref(self) -> Value {
```
vgao1996, line 572: I feel like this function (or any function that dereferences a CheckedRef) should be marked as unsafe, but I'm not 100% sure since this is an internal API.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -607,46 +568,24 @@ impl Value {
...
+                CheckedRef::Container(c) => ValueImpl::Container((&*c).clone()),
+            }
+        };
+        Value(v)
     }
 }
 
 impl Reference {
-    pub fn read_ref(self) -> PartialVMResult<Value> {
+    pub fn read_ref(self) -> Value {
```
vgao1996, line 588: This should be unsafe fn -- you are dereferencing a raw pointer under the hood.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -511,85 +537,20 @@ impl Container {
...
-            (Vec(_), _)
-            | (VecU8(_), _)
-            | (VecU64(_), _)
-            | (VecU128(_), _)
-            | (VecBool(_), _)
-            | (VecAddress(_), _) => {
-                return Err(PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR)
-                    .with_message(format!("cannot compare references {:?}, {:?}", self, other)))
+impl ReferenceImpl {
+    pub(crate) fn equals(&self, other: &ReferenceImpl) -> PartialVMResult<bool> {
```
vgao1996, line 541: This (and `Value::equals`) should probably be unsafe.

vgao1996 (commented):
Still need to take a closer look, but here are some early feedbacks regarding unsafe fn

language/move-vm/types/src/values/values_impl.rs
tnowacki: Yeah its a tricky one... I suppose I should... but it will make lots of other things unsafe.

language/move-vm/types/src/values/values_impl.rs
tnowacki: I'll see how it looks doing that

language/move-vm/types/src/values/values_impl.rs
tnowacki: yeah lots of things will be with this change, basically anything that reads/writes to refs. Which is a lot of things

tnowacki:
> Still need to take a closer look, but here are some early feedbacks regarding unsafe fn.
> 
> I think one thing to point out is that `write_safe` is probably not the only thing that's unsafe -- `read_ref` and more broadly, anything that dereferences a CheckedRef (a.k,a, raw pointers) are also unsafe. Consider this case:
> 
> ```
> let locals = Locals::new(4);
> 
> locals.store_loc(0, Value::u64(0));
> locals.store_loc(1, Value::u64(0));
> 
> let r1 = locals.borrow_loc(0).unwrap();
> let r2 = locals.borrow_loc(1).unwrap();
> 
> std::mem::drop(locals);
> 
> r1.equals(r2);
> ```
> 
> I'm aware this is a terrible way to use the value APIs, but this is exactly the point of marking them as unsafe -- forcing the Move VM to annotate its usages, making it easier for us to review them and make sure they are indeed doing sane things.

Very perceptive :) I saw exactly this after working on #175, seems you just saw it a bit earlier than myself

vgao1996:
Not sure why the hardhat-tests are failing... Could you try to rebase?

vgao1996 (approved):
LGTM, but let's leave some time for @wrwg to have a look, just to make sure he's on board with the general direction.

language/extensions/async/move-async-vm/src/natives.rs
```diff
@@ -75,7 +75,7 @@ fn native_send(
     let ext = context.extensions_mut().get_mut::<AsyncExtension>();
     let mut bcs_args = vec![];
     while args.len() > 2 {
-        bcs_args.push(pop_arg!(args, Vector).to_vec_u8()?);
+        bcs_args.push(pop_arg!(args, Vec<u8>));
```
wrwg, line 78: Ah! That actually works, great. Problem with that kind of magic is that the APIs aren't clearly documented and buried somewhere in all those trait impls.

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -395,12 +395,9 @@ fn native_add_box(
...
     let mut table_data = table_context.table_data.borrow_mut();
 
+    // see REFERENCE SAFETY EXPLANATION in values
     let val = args.pop_back().unwrap();
-    let key = args
-        .pop_back()
-        .unwrap()
-        .value_as::<Reference>()?
-        .read_ref()?;
+    let key = unsafe { args.pop_back().unwrap().value_as::<Reference>()?.read_ref() };
```
wrwg, line 400: Wondering: can't we isolate the `unsafe` somewhere in the value impl? Really don't like to see this in user code.

language/move-vm/types/src/values/mod.rs
```diff
@@ -1,6 +1,18 @@
...
+//! REFERENCE SAFETY EXPLANATION
+//! References in Move are proved safe statically in the reference safety check
+//! (bytecode_verifier/reference_safety). This check guarantees that there are no dangling
+//! references and no memory leaks. But the access pattern for reading/writing to Values must
+//! abide by these safety rules (which the VM interpreter does. You should refrain from writing to
+//! reference values outside of the VM or native functions (with special care))
+//! However, it is **very important** that any instance of `Locals` is not dropped until after
+//! all Values are consumed. If this is not followed, reading/writing to Values *will* read from
+//! invalid pointers.
+
```
wrwg, line 15: I think this is kind of bummer for usage of the VM. We want to use references in adapters (doesn't Sui does so) and custom embeddings of the VM, as well as native functions implementations, avoiding the need for using `unsafe` all over the code. Unsafe Rust should be isolated. Can you wrap the unsafe usage somewhere inside the VM?

wrwg (changes requested):
Concerned about the spread of `unsafe` across the code.

@sblackshear @davidiw

tnowacki:
> I think this is kind of bummer for usage of the VM. We want to use references in adapters (doesn't Sui does so) and custom embeddings of the VM, as well as native functions implementations, avoiding the need for using unsafe all over the code. Unsafe Rust should be isolated. Can you wrap the unsafe usage somewhere inside the VM?

Anything with references in this implementation will be inherently unsafe. Even reading from them could be unsafe, if an improper write occurs. So there isn't really way to encapsulate the logic, which we already do for adapters, could do some endpoints, and can't really do for native functions:

On native functions: 
And there really isn't anything that can be done from the native function side of things, since native functions get references as inputs (and the native function implementation is not bound by Move's safety rules, the implementation could be wrong). Native functions already could have messed this up, but it would have been a graceful invariant violation. Now it will be some deeply disturbing, silent pointer failure. Which is potentially an issue. However, to me this doesn't feel like an issue in the sense that native functions are stating "I am extending the VM runtime", and as such you need to be just as safe as the VM runtime.

On the adapter:
None of the safety concerns are exposed to the adapter. `Value` is only used inside of the runtime. The adapter talks to the VM with BCS serialized values, both for inputs and outputs. So this "wrapping" is already done indirectly by the usage of serialized values. 

For other endpoints (testing infra and such):
Its a bit annoying, and I could try to split the `Value` enum into reference and non reference. It would bloat the API but it would clean up at least a few endpoints... maybe...

language/extensions/async/move-async-vm/src/natives.rs
tnowacki: agreed its a bit hard to discover, not sure if @vgao1996 has any thoughts on that (to be clear the VMValueCast stuff has been around a while, not new to this PR)

wrwg:
Just a further note, not actionable but related to this PR.

If we want to have a more efficient implementation of values, I think we could go a different approach. What I sketch is motivated both by implementing the Move memory model on top of the EVM and for the Prover.

(a) Linear Memory

Borrow semantics is a model for linear addressable RAM, including pointer indirection and address arithmetic, and I think will be most efficiently implemented by reflecting those aspects. So why not represent any value as a byte-addressable region of memory. You can borrow offsets in this region, read values, and mutate them. An unsafe, untyped value representation would then be `Value(*[u8])`.

The lifetime of the region the value points to is guaranteed to be valid by our own implementation of memory allocation. We can also skip the GC (as Solidity does) assuming that transactions are short running. I.e. we use arena style allocation for `*[u8]`. The stack can still be shrinking and growing, but once allocated stack memory will stay in the VMs possession.  In any case we can assume that `*[u8]` is always Rust-safe but possibly Move-unsafe, as we never give memory back to Rust.

As to how we layout memory,  this was investigated as part of EVM Move. Assuming a field in a struct has a fixed offset, field borrowing simply amounts to `Value(r) => Value(r[offs..offs+size])`. Similar, borrowing a local amounts to slicing the stack frame, For globals we can give them their own region.

Nested structs can be inlined, or represented via indirections like vectors, which inherently need indirections because they have no fixed size.

(b) Generics and Variants

We can avoid enums for different value variants all together. A `Value(*[u8])` can be only interpreted by providing a type. This type is either statically known to the program (no generics), or passed as a parameter to a function (with generics).

Since `*[u8]` always points to valid Rust memory and has a given size, we cannot destroy Rust integrity by using the wrong type when interpreting a value, including when mutating it.

There is one exception regards integrity, that is when we have an indirection to a vector (or struct if we choose so) embedded in `*[u8]`. In this case we must create a new `Value` from the data inside the region, which might be corrupted if we have the wrong type to interpret it.  There are multiple ways how to deal with this:

- Trust bytecode verification for internal VM code, and assume type soundness. For the external API, use `ExtValue(TypeTag, Value)`
- Validate the indirection to be well-defined. As we know all memory we ever allocated, we can check whether the `Value(*[u8])` constructed from the content of some other value actually points to a valid region. Valid means in the Rust sense, it can be still invalid in the Move sense.

PR #164, opened 2022-03-30
[evm] Created a hardhat project for Move contracts

junkil-park:
This PR creates a hardhat project for Move-on-EVM contracts. In the `contracts` directory, two contracts are written in Move:
- FortyTwo: a stateless contract
- Greeter: a contract with state

Thanks to Victor, this project has an automatic script (`compile_move_and_test.sh`) to compile and test Move contracts in hardhat. Currently, not all tests pass due some unsupported features of `move-to-yul` (e.g., constructor translation, returning vectors, ...). These contracts (and more to be added later) can be used as running examples for our Move-hardhat integration work.

The `contracts` directory also contains equivalent Solidity contracts in parallel. An automatic script (`compile_sol_and_test.sh`) will compile and test those Solidity contracts.

## Motivation

test Move contracts in hardhat

yes

## Test Plan

compile_move_and_test.sh
compile_sol_and_test.sh


language/evm/hardhat-examples/.gitignore
```diff
@@ -0,0 +1,9 @@
+node_modules
+.env
+coverage
+coverage.json
+typechain
+
+#Hardhat files
+cache
+artifacts
```
vgao1996, line 9: I like the way you handled this! I was thinking about checking in the whole hardhat project but was wary of including the node packages and other stuff, and the .gitignore approach seems to address this well.

language/evm/hardhat-examples/contracts/FortyTwo.move
```diff
@@ -0,0 +1,26 @@
+// FortyTwo.move
```
vgao1996, line 1: You can probably drop this comment -- I included it only because it was part of the tutorial and I needed to tell the reader which file this was.

language/evm/hardhat-examples/test/FortyTwo.test.js
```diff
@@ -0,0 +1,25 @@
+const { expect } = require("chai");
```
vgao1996, line 1: Right now, each Move contract and its Solidity equivalent share the same name and are used by the same test script. I don't know if this is a good thing or not...

vgao1996 (approved):
LGTM, but you'll need to fix the lints.

language/evm/hardhat-examples/contracts/FortyTwo.move
junkil-park: done.

language/evm/hardhat-examples/test/FortyTwo.test.js
junkil-park: The upside of the shared test script is that we maintain only one test script per contract. Moreover, it shows that a Move contract can replace a Solidity contract, and they are interchangeable. The downside is that some hidden "build cache" affects the workflow, so `npx hardhat clean` is needed before each "compile and test". It's included in the scripts already. Moreover, we may want a scenario in the future that Move contracts and Solidity contracts coexist in the same hardhat project, so that they cross-contract-call each other.

rahxephon89 (dismissed):

PR #179, opened 2022-04-04
[draft][evm] allow move package build to target the EVM architecture

vgao1996:
This makes it possible to compile Move into EVM bytecode using the package system.


jolestar:
How about the dependencies? EVM bytecode package should include all dependencies, but Move does not need to do that.

language/tools/move-cli/src/package/cli.rs
```diff
@@ -86,7 +139,10 @@ pub enum PackageCommand {
     },
     /// Build the package at `path`. If no path is provided defaults to current directory.
     #[clap(name = "build")]
-    Build,
+    Build {
+        #[clap(long = "arch", parse(try_from_str = Architecture::try_parse_from_str))]
```
wrwg, line 143: So there is already a new option to the package system and compiler as seen here: https://github.com/diem/move/blob/main/language/extensions/async/examples/account/Move.toml#L11.

It currently can have values "async" and "evm". This is passed into the compiler in order to enable attribute processing. For EVM, for example, `#[callable] fun f` leads to generation of native functions `#[extern] public native fun call_f` for cross-contract calls. I wonder whether we should try to  unify this.

vgao1996:
@jolestar you are right that an EVM contract includes all of its dependencies. When we compile Move into EVM bytecode, we need to follow the same rule. Let's say you have two modules A, B and A depends on B. Then when you run `move-cli`
in A, it will generate a contract containing code from both A and B.

It should be noted however, the current behavior that only one single contract is generated is tentative. I think we'll likely allow users to specify multiple contracts using attributes. We already have an attribute for this ("#[contract]") but it's unused.

wrwg (dismissed):

PR #181, opened 2022-04-04
stub `impl TableResolver` for DeltaStorage

msmouse:
## Motivation

To unblock downstream tests.

In Aptos repo we are trying to enforce the table extension to be installed everywhere, and this not supporting tables breaks compilation.

y

## Test Plan

Downsteam tests pass

### DIFF
--- a/language/move-vm/test-utils/src/storage.rs
+++ b/language/move-vm/test-utils/src/storage.rs
@@ -111,6 +111,33 @@ impl<'a, 'b, S: ResourceResolver> ResourceResolver for DeltaStorage<'a, 'b, S> {
     }
 }
 
+#[cfg(feature = "table-extension")]
+impl<'a, 'b, S: TableResolver> TableResolver for DeltaStorage<'a, 'b, S> {
+    fn resolve_table_entry(
+        &self,
+        handle: &TableHandle,
+        key: &[u8],
+    ) -> std::result::Result<Option<Vec<u8>>, Error> {
+        // TODO: No support for table deltas
+        self.base.resolve_table_entry(handle, key)
+    }
+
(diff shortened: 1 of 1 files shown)


language/move-vm/test-utils/src/storage.rs
```diff
@@ -111,6 +111,33 @@ impl<'a, 'b, S: ResourceResolver> ResourceResolver for DeltaStorage<'a, 'b, S> {
...
 
+#[cfg(feature = "table-extension")]
+impl<'a, 'b, S: TableResolver> TableResolver for DeltaStorage<'a, 'b, S> {
+    fn resolve_table_entry(
+        &self,
+        handle: &TableHandle,
+        key: &[u8],
+    ) -> std::result::Result<Option<Vec<u8>>, Error> {
+        // n.b. Doesn't support table delta for now
+        self.base.resolve_table_entry(handle, key)
```
vgao1996, line 122: The motivation sounds reasonable to me, and I'd love to see TableResolver being implemented for all of these mock storages. With that being said, I wonder if it would be better to leave these as `unimplemented!`? (I could be wrong though as I don't know exactly how you are planning to use this in its incomplete state.)

msmouse: The `DeltaStorage` is used in the txn replay tool, with `self.delta` "patched" upon `self.base`. So this should work as expected, lacking the ability to patch table related change set.

davidiw (approved):
As tables is an experimental feature and the author is currently out of town, @wrwg , accepting to unblock. Let's follow up with him when he gets back if he has concerns on the lack of delta support -- as well as the timeline for when delta support may be required.

davidiw (dismissed):
guess approvals are dismissed on this repo on fixups ;)

PR #190, opened 2022-04-06
[evm] add abort_with string

emmazzz:
## Motivation

This PR adds `abort_with` native function that allows Move code to abort with a custom error message. This works for the truffle-style tests (in Revert.truffle.test.js) but not the hardhat test (in Revert.test.js).


junkil-park:
Thanks! LGTM. Let me land it and test it.

junkil-park (dismissed):

PR #194, opened 2022-04-06
RFC: UnsafeStaticTableResolver

msmouse:
## Motivation

Because `pub trait Any: 'static` must be static, there's no way to create the `NativeTableContext` in production environment where the resolver has a shorter lifetime. 

This unsafe piece of code is the best I can come up with for now. Let me know how people envision us to proceed.

Also:
* Connected unit tests with `InMemoryStorage`.
* Fixed that `length()` queries storage for newly created tables in the current txn. (unbroken test with `InMemoryStorage`)
* Fixed that global_value() didn't consider deleted item in mem and didn't cache None's from the resolver. (added unit tests)

Y

## Test Plan

Existing coverage, and downstream tests

## Related PRs

https://github.com/diem/move/pull/150

### DIFF
--- a/language/extensions/move-table-extension/src/lib.rs
+++ b/language/extensions/move-table-extension/src/lib.rs
@@ -80,6 +80,49 @@ pub trait TableResolver {
     ) -> InternalGasUnits<GasCarrier>;
 }
 
+struct UnsafeStaticTableResolver {
+    inner: *const dyn TableResolver,
+}
+
+impl UnsafeStaticTableResolver {
+    fn unsafe_new(resolver: &dyn TableResolver) -> Self {
+        unsafe {
+            Self {
+                inner: std::mem::transmute::<&dyn TableResolver, *const dyn TableResolver>(
+                    resolver,
+                ),
+            }
+        }
+    }
+
+    fn unsafe_get(&self) -> &dyn TableResolver {
+        unsafe { &*self.inner as &dyn TableResolver }
+    }
+}
+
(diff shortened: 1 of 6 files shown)


language/extensions/move-table-extension/src/lib.rs
```diff
@@ -80,6 +82,49 @@ pub trait TableResolver {
     ) -> InternalGasUnits<GasCarrier>;
 }
 
+struct UnsafeStaticTableResolver {
```
zekun000, line 85: you probably want to add 50 lines of comments explaining the safety contract of this thing and how to use it properly

msmouse: Yeah I will want to. When I think people will agree on this solution..

wrwg:
Hey, it's unfortunate that we need `'static` because of the `Any` trait; I've not thought about this when designing the extension model, which I previously only used with pure, reference free data. Are you in a hurry to land this? Otherwise I can take a closer look first how (and if) we can refactor the extension model to avoid this.

msmouse:
@wrwg  Thanks. It's not THAT urgent. The inherent limitation of rust is downcasting can only happen on `'static`.  We either use this or a similar hack on the extension side (and document 50 lines to explain the implications), or ask the user (VM adaptor) to provide only `'static` things, which will very likely to force the user to write some unsafe code (like in the case of Aptos, because the resolver is actually the state_view, which have a non-static lifetime.)

PR #195, opened 2022-04-07
[vm] Added structured stack trace data to VMError for cleaner diagnostic presentation

awelc:
## Motivation

Prior to this PR we were embedding VM crash info into the `VMError` which resulted in the execution state information displayed on assertion failures and explicit aborts not being super-user friendly.

In this PR we add structured  stack trace data to `VMError` instead and customize the display to better match the usual/expected way of displaying stack traces.

Consequently, what we have now looks as follows:

stack_after

Instead of before:

stack_before

Yes


language/tools/move-unit-test/src/test_reporter.rs
```diff
@@ -219,6 +226,87 @@ impl TestFailure {
...
+        let end_line_index = files.line_index(*id, loc.end() as usize)?;
+        let end_line_number = files.line_number(*id, end_line_index)?;
+        if start_line_number == end_line_number {
+            Ok(start_line_number.to_string())
+        } else {
+            Ok(format!("{}-{}", start_line_number, end_line_number))
+        }
+    }
+
+    fn report_exec_state(test_plan: &TestPlan, exec_state: &ExecutionState) -> String {
```
sblackshear, line 257: I'm not sure how hard it is to implement this in a way that does not depend on `TestPlan` (and definitely not needed for this PR). But if we could make this a function of `ExecutionState` + `Option<SourceMap>` (or some trait that `TestPlan` could implement), I could definitely anticipate other tooling benefiting from the friendlier stack traces (e.g., `move sandbox run` in the CLI).

sblackshear (dismissed):

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -570,6 +570,21 @@ impl Interpreter {
...
+        for frame in self.call_stack.0.iter().rev() {
+            // collect frames in the reverse order as this is what is
+            // normally expected from the stack trace (outermost frame
+            // is the last one)
+            stack_trace.push((
+                frame.function.module_id().cloned(),
+                frame.function.index(),
+                frame.pc,
+            ));
+        }
```
tnowacki, line 585: nit, `self.call_stack.0.iter().rev().map(...).collect()`

language/tools/move-unit-test/src/test_reporter.rs
```diff
@@ -219,6 +226,87 @@ impl TestFailure {
...
+            buf.push_str("stack trace\n");
+            let mut files = SimpleFiles::new();
+            let mut file_mapping = HashMap::new();
+            for (fhash, (fname, source)) in &test_plan.files {
+                let id = files.add(*fname, source.as_str());
+                file_mapping.insert(*fhash, id);
+            }
+
+            for frame in stack_trace {
+                if let Some(module_id) = &frame.0 {
```
tnowacki, line 270: nit, match over if-let-else

language/tools/move-unit-test/src/test_reporter.rs
```diff
@@ -219,6 +226,87 @@ impl TestFailure {
...
+                    if let Some(named_module) = test_plan.module_info.get(module_id) {
```
tnowacki, line 271: nit, match over if-let-else

language/tools/move-unit-test/src/test_reporter.rs
```diff
@@ -219,6 +226,87 @@ impl TestFailure {
...
+                        if let Ok(function_source_map) =
```
tnowacki, line 272: nit match over if-let-else

language/tools/move-unit-test/src/test_reporter.rs
```diff
@@ -219,6 +226,87 @@ impl TestFailure {
...
+                                    "\t{}::{}({}:{})\n",
+                                    module_id.name(),
+                                    fn_name,
+                                    file_name,
+                                    Self::get_line_number(&loc, &files, &file_mapping)
+                                )
+                                .to_string(),
+                            );
+                        } else {
+                            return "\tmalformed stack trace (no source map)".to_string();
```
tnowacki, line 297: Actually, since you are returning in all of these you should do 
```
let function_souce_map = match named_module.source_map.get_function_source_map(frame.1) {
  Some(function_souce_map) => function_souce_map,
  None => return "\tmalformed stack trace (no source map)".to_string();
};
```

language/tools/move-unit-test/src/test_reporter.rs
awelc: I have been staring at this code for a while and I don't quite see a nice and generic way of refactoring this, mostly because we need quite a bit of info that could be conceivably obtained in different ways, for example:
- source map and compiled module (of course)
- mapping from file hash to file name
- `SimpleFiles` instance to get info on byte locations of line starts

I am not saying that the refactoring is impossible, I just think that it should happen once we have a concrete other place where we want to reuse this functionality, rather than making it proactively.

sblackshear (approved):
/land

PR #197, opened 2022-04-07
[move-vm] Expose more function and type APIs in the session

tnowacki:
## Motivation

- Exposing more access to the cache gives adapters more flexibility
- Allows adapters to implement checks over cached data (without having to make their own cache)

## Test Plan

- eyes. endpoints are already tested, just now expoed

### DIFF
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -4,6 +4,7 @@
 use crate::{
     logging::expect_no_verification_errors,
     native_functions::{NativeFunction, NativeFunctions},
+    session::LoadedFunctionInstantiation,
 };
 use move_binary_format::{
     access::{ModuleAccess, ScriptAccess},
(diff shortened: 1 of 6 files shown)


language/move-vm/runtime/src/loader.rs
```diff
@@ -1344,7 +1345,7 @@ impl<'a> Resolver<'a> {
 pub(crate) struct Module {
     id: ModuleId,
     // primitive pools
-    module: CompiledModule,
+    module: Arc<CompiledModule>,
```
sblackshear, line 1348: My only questions is whether this is likely to have unpleasant perf implications.

sblackshear (approved):

language/move-vm/runtime/src/loader.rs
tnowacki: I don't think so. `Module` is already wrapped in an `Arc`, if anything having this boxed might improve things... as the `Module` size is significantly smaller

PR #199, opened 2022-04-08
[account-address] Input and output with a 0x for hex output

gregnazario:
Accepts 0x as an input, and will print the output as a 0x for AccountAddress.
This removes confusion on whether it's in a different encoding, and keeps
it simple and backwards compatible.

### DIFF
--- a/language/move-core/types/src/account_address.rs
+++ b/language/move-core/types/src/account_address.rs
@@ -68,7 +68,7 @@ impl AccountAddress {
 
         let hex_len = literal.len() - 2;
 
-        // If the string is too short, pad it
+        // If the string is too short, pad it because it needs to be exactly the right number of bytes
         if hex_len < Self::LENGTH * 2 {
             let mut hex_str = String::with_capacity(Self::LENGTH * 2);
             for _ in 0..Self::LENGTH * 2 - hex_len {
@@ -92,7 +92,7 @@ impl AccountAddress {
     }
 
     pub fn to_hex(&self) -> String {
-        format!("{:x}", self)
+        self.short_str_lossless()
     }
 
     pub fn from_bytes<T: AsRef<[u8]>>(bytes: T) -> Result<Self, AccountAddressParseError> {
@@ -206,7 +206,7 @@ impl From<&AccountAddress> for [u8; AccountAddress::LENGTH] {
 
 impl From<&AccountAddress> for String {
     fn from(addr: &AccountAddress) -> String {
-        ::hex::encode(addr.as_ref())
+        addr.to_hex_literal()
     }
 }
 
@@ -214,7 +214,7 @@ impl TryFrom<String> for AccountAddress {
     type Error = AccountAddressParseError;
 
     fn try_from(s: String) -> Result<AccountAddress, AccountAddressParseError> {
-        Self::from_hex(s)
+        AccountAddress::from_str(&s)
     }
 }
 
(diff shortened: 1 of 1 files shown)


bmwill:
cc @tnowacki since I think 'hex_literal' may mean something specific in the move source language

gregnazario:
> cc @tnowacki since I think 'hex_literal' may mean something specific in the move source language

Please let me know, I'm just trying to fix the general usability of this, since it's so tightly bound to must be full digits in hex, without a `0x` and without any missing characters, e.g. `0x1`.  or just `1`.

I'm trying to make CLI inputs more consistent at the source, rather than having to build intricate systems around it

tnowacki:
If you are trying to fix CLI inputs, or anything else for that matter, to make it more consistent with the source language, I would just use the source language's parsing. It is exposed here:
https://github.com/diem/move/blob/180a066094704b2411d3970758b2a05d2e406213/language/move-compiler/src/shared/mod.rs#L67
Which should be, `move_compiler::shared::parse_address`

tnowacki (changes requested):
Let's try to just use the source language API where you want it, best not to change the parsing in core types

tnowacki:
Keep in mind, parse_address parses to decimal if no 0x prefix is found. Basically it treats the address parsing as if it was any other integer value

gregnazario:
> If you are trying to fix CLI inputs, or anything else for that matter, to make it more consistent with the source language, I would just use the source language's parsing. It is exposed here:
> 
> https://github.com/diem/move/blob/180a066094704b2411d3970758b2a05d2e406213/language/move-compiler/src/shared/mod.rs#L67
> 
> 
> Which should be, `move_compiler::shared::parse_address`

I'm not trying to just fix CLI inputs, I'm trying to fix the general experience around inputs and outputs of move addresses.  This is just to separate out 2 problems:

1. If you don't have exactly the number of digits, it'll not work, which the 0x hex literal parser already accomplished.
2. Anywhere that uses this output has to be changed in or out.  Serialization outputs end up being different than hex string outputs, which can also be confusing.

I'm open to other ideas, but I'd like us to have a way where it's straight forward that it's a hex string, not just a bunch of letters, and that it can be transported around easily along with allowing for things like 0x1 instead of 0x00000000000000000001.

Also, I'm curious.  Why does `parse_address` default to decimal?

tnowacki:
> Also, I'm curious. Why does parse_address default to decimal?

Thats how addresses work in the source language.
```
let a1: address = @66;
let a2: address = @0x112;
assert!(a1 == a2);
```
This is to stay consistent with the other numerical syntax in move
```
let u1 = 66;
let  u2 = 0x112;
assert!(a1 == a2);
```
This works for any integer type, `u8`, `u64`, or `u128`

gregnazario:
> > Also, I'm curious. Why does parse_address default to decimal?
> 
> Thats how addresses work in the source language.
> 
> ```
> let a1: address = @66;
> let a2: address = @0x112;
> assert!(a1 == a2);
> ```
> 
> This is to stay consistent with the other numerical syntax in move
> 
> ```
> let u1 = 66;
> let u2 = 0x112;
> assert!(a1 == a2);
> ```
> 
> This works for any integer type, `u8`, `u64`, or `u128`

Okay, so what that's telling me is that addresses, how they are printed now in text must have the `0x` before it to be proper.  I can remove the optionality of converting from string if it doesn't have a 0x.

tnowacki:
> Okay, so what that's telling me is that addresses, how they are printed now in text must have the 0x before it to be proper. I can remove the optionality of converting from string if it doesn't have a 0x.

Where are you removing this optionality?

gregnazario:
Closing for https://github.com/move-language/move/pull/7

PR #200, opened 2022-04-08
[sccache] Fix sccache folders to be consistent and work with move repo

gregnazario:
Sccache isn't setup correctly, so this'll make the folders consistent.

### DIFF
--- a/x.toml
+++ b/x.toml
@@ -9,8 +9,8 @@ prefix = "sccache/diem/"
 public = true
 region = "us-west-2"
 endpoint = "https://s3-us-west-2.amazonaws.com"
-required-cargo-home = "/opt/cargo/"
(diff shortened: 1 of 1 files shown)


bmwill:
Do we still use sccache? I seemed to remember us having some issues with it a while back and can't remember if its even still used

gregnazario:
> Do we still use sccache? I seemed to remember us having some issues with it a while back and can't remember if its even still used

We can remove it, this just is to make it consistent if some one still uses it

bmwill:
/help

Guess i'll join in the fun

gregnazario:
Closing, as there doesn't seem to be a want to fix the sccache config and I no longer have access to this repo

PR #208, opened 2022-04-11
[move-lang][vm] Allow non-`'static` native context extensions.

wrwg:
This uses the `Tid<'a>` type from the [better_any](https://docs.rs/better_any/latest/better_any/) crate which acts as a replacement for the standard `Any` type in order to implement `NativeContextExtensions`. This allows the polymorphic values stored in the extensions to be non-`'static`, which `Any` does not.

As a consequence, it is now `NativeContextExtensions<'a>`, where `'a` is placeholder for the lifetime of any references stored in some extensions. Only one lifetime parameter is allowed for such extensions because of restrictions of `better_any`.

A user of extensions should not see any of this, however, structures describing extensions must use `#[derive(Tid)]`, and any type parameters those have must do as well. This is so the crate can compute a `TypeId` for such types.

The `NativeContextExtensions<'a>` type has been moved in its own module, and a test has been added to verify that this new feature works.

## Motivation

Allow extensions to contain non-static references, as e.g. for #194 

Yes

## Test Plan

Added test

### DIFF
--- a/language/move-vm/runtime/src/native_extensions.rs
+++ b/language/move-vm/runtime/src/native_extensions.rs
@@ -1,0 +1,82 @@
+// Copyright (c) The Diem Core Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use better_any::{Tid, TidAble, TidExt};
+use std::{any::TypeId, collections::HashMap};
+
+/// A data type to represent a heterogeneous collection of extensions which are available to
+/// native functions. A value to this is passed into the session function execution.
+///
+/// The implementation uses the crate `better_any` which implements a version of the `Any`
+/// type, called `Tid<`a>`, which allows for up to one lifetime parameter. This
+/// avoids that extensions need to have `'static` lifetime, which `Any` requires. In order to make a
+/// struct suitable to be a 'Tid', use `#[derive(Tid)]` in the struct declaration. (See also
+/// tests at the end of this module.)
+#[derive(Default)]
+pub struct NativeContextExtensions<'a> {
+    map: HashMap<TypeId, Box<dyn Tid<'a>>>,
+}
+
+impl<'a> NativeContextExtensions<'a> {
+    pub fn add<T: TidAble<'a>>(&mut self, ext: T) {
+        assert!(
+            self.map.insert(T::id(), Box::new(ext)).is_none(),
+            "multiple extensions of the same type not allowed"
+        )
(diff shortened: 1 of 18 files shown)


wrwg:
@tnowacki not sure whether I need to bother you as a reviewer for all such changes, so just mentioning, but feel free to jump in.

msmouse (approved):
Works for me. Thanks.

language/move-vm/runtime/src/native_extensions.rs
```diff
@@ -0,0 +1,82 @@
...
+/// struct suitable to be a 'Tid', use `#[derive(Tid)]` in the struct declaration. (See also
+/// tests at the end of this module.)
+#[derive(Default)]
+pub struct NativeContextExtensions<'a> {
+    map: HashMap<TypeId, Box<dyn Tid<'a>>>,
+}
+
+impl<'a> NativeContextExtensions<'a> {
+    pub fn add<T: TidAble<'a>>(&mut self, ext: T) {
+        assert!(
```
tnowacki, line 22: should this maybe return a `VMResult`? That way we can handle the assert gracefully?

I know we don't do this everywhere in the VM, but we try to do it where we can (I hope)

tnowacki: And it shouldn't be that much more painful for anyone using this, they will likely just have to slap a `?` on the call

tnowacki (commented):
VM changes look fine to me, I have some concerns about the asserts, but I don't think they are new to this PR so we could fix them later

language/move-vm/runtime/src/native_extensions.rs
wrwg: That's actually an interesting thing we should discuss. At Google I learned that fail fast and assertive in code which correctness does not depend on user input is preferred. So I would say that

- It should be VMResult whenever the behavior depends e.g. on bytecode input.
- It should be actually assert if it does not depend on bytecode. (This is also easier to debug BTW.)

We can continue to discuss this. Landing for now.

/land

wrwg:
> @wrwg ❗ This PR is still missing approvals, unable to queue for landing

@msmouse can you land this for me? All we folks of Meta seem to have been removed from the Diem org, we cannot land anymore, nor be added as reviewers... At least I guess that is what the error comes from, because this PR is approved.

PR #264, opened 2022-11-24
[move-to-yul] Add supports for external calls with try/catch behaviors + bug fix#2_3

sahithiacn:
## **Motivation**

1.As stated [here](https://docs.soliditylang.org/en/v0.8.13/control-structures.html?highlight=try#try-catch), Solidity supports different kinds of catch blocks depending on the type of error (Error(string memory reason), Panic(uint errorCode), and bytes memory lowLevelData). To describe the behavior of try/catch statements, a struct type ExternalResult is proposed:

```
    struct ExternalResult<T> has copy, drop, store {
        value: Option<T>,
        err_vec: Option<vector<u8>>,
        err_str: Option<vector<u8>>,
        panic_code: Option<U256>
    }
```
Users can specify an external function like this
```
    #[external(sig=b"foo() returns (uint)")]
    public native fun foo(contract: address): ExternalResult<U256>;
```

By using the APIs provided by ExternalResult, the return value or error information can be tested and/or extracted:
```
    #[callable, pure]
    public fun call_foo(addr: address): U256 {
        let v = foo(addr);
        if (ExternalResult::is_ok(&v)) {
            ...
        } else if (ExternalResult::is_err_reason(&v)) {
            ...
        } else if (ExternalResult::is_panic(&v)) {
            return ExternalResult::unwrap_panic(v);
        };
        ...
    }
```
2.Fix a bug of the method Vector::pop_back for a vector that stores a non-value type in memory.

Yes

## **Test Plan**

TestExternalResult.move

## **Related PRs**

https://github.com/diem/move/pull/140


dhaneshacn (approved):

dhaneshacn (dismissed):

dhaneshacn (dismissed):

language/move-prover/bytecode/src/well_formed_instrumentation.rs
```diff
@@ -35,7 +35,7 @@ impl WellFormedInstrumentationProcessor {
         Box::new(Self {})
     }
 }
-
+#[allow(clippy::redundant-clone)]
```
c0wjay, line 38: It looks like there's a typo error, and it seems that `#[allow(clippy::redundant_clone)]` is correct.

PR #267, opened 2022-12-02
[move-lang][vm] Allow non-'static native context extensions. (#3)_4

sahithiacn:
## Motivation
Allow extensions to contain non-static references

This uses the Tid<'a> type from the [better_any](https://docs.rs/better_any/latest/better_any/) crate which acts as a replacement for the standard Any type in order to implement NativeContextExtensions. This allows the polymorphic values stored in the extensions to be non-'static, which Any does not.

As a consequence, it is now NativeContextExtensions<'a>, where 'a is placeholder for the lifetime of any references stored in some extensions. Only one lifetime parameter is allowed for such extensions because of restrictions of better_any.

A user of extensions should not see any of this, however, structures describing extensions must use #[derive(Tid)], and any type parameters those have must do as well. This is so the crate can compute a TypeId for such types.

The NativeContextExtensions<'a> type has been moved in its own module, and a test has been added to verify that this new feature works.

Yes

## Test Plan

Modified  tests under integration,unit testsuite

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,6 +148,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "better_any"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b359aebd937c17c725e19efcb661200883f04c49c53e7132224dac26da39d4a0"
+dependencies = [
+ "better_typeid_derive",
+]
+
+[[package]]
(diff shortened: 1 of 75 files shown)


KoushikGavini:
Hi Sahiti, 

Approving conditionally, can we add better description for test coverage.

dhaneshacn (dismissed):

